PR #1, opened 2021-06-09
ImageNet preprocessing functions

AdityaKane2001:
Added

1.  Random sized crop
2.  Scale and center crop
3. PCA jitter


AdityaKane2001:
@sayakpaul 
Adding protection to main branch to avoid accidental commits. Requires at least one review approval.

sayakpaul:
@AdityaKane2001 please also add @MorganR as a reviewer.

dataset/imagenet.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+  def __init__(self,tfrecs_filepath = None,
+               batch_size = 128,
+               image_size = 224,
+               augment_fn = 'default'):
+
+    if tfrecs_filepath is None:
+      raise ValueError('List of TFrecords paths cannot be None')
+    print('hi')
```
sayakpaul, line 34: Is it for debugging?

dataset/imagenet.py
```diff
@@ -0,0 +1,148 @@
...
+    self.tfrecs_filepath = tfrecs_filepath
+    self.batch_size = batch_size
+    self.image_size = image_size
+    self.augment_fn = augment_fn
+
+  
+  def _read_tfrecs(self):
```
sayakpaul, line 41: Would suggest adding docstrings for all cases.

dataset/imagenet.py
```diff
@@ -0,0 +1,148 @@
...
+          'image' : image,
+          'height' : self.image_size,
+          'width' : self.image_size,
+          'filename' : example['filename'],
+          'label' : example['label']
+        }
+  
+  @tf.function
+  def _pca_jitter(self, image):
+      image = tf.cast(image, tf.float32)/255.
```
sayakpaul, line 128: Instead of placing this scaling step here, I would recommend placing it in the function that will be used by all kinds of datasets i.e. train, validation, and test.

dataset/imagenet.py
```diff
@@ -0,0 +1,148 @@
...
+ 
+  def make_dataset(self, ):
+    ds = self._read_tfrecs()
+
+    if self.augment_fn == 'default':
+      ds = ds.map(lambda example: self._random_sized_crop(example))
+
+    else:
+      ds = ds.map(lambda example: self.augment_fn(example))
+    return ds
```
sayakpaul, line 148: If my understanding is correct, the current augmentation pipeline only consists of random resized crops, right? But should that be the case?

Also, instead of naming this script as `imagenet.py` I would recommend adding a bit more description. Something like `imagenet_tfrec_loader.py` since this script seems to contain utilities for parsing the TFRecords and creating a `tf.data.Dataset` out of them.

sayakpaul (changes requested):
@AdityaKane2001 looks great for a first PR. Here are some additional suggestions (apart from the ones I have commented):

* I did not see a script for creating the TFRecords. Could you help point me to it? 
* I would suggest using a linter like Black.
* I would also like to see a couple of image batches after parsing the TFRecords. A notebook or a demo script might be good to test the data loaders out.

dataset/imagenet.py
AdityaKane2001: Yes, I'll remove it in next commit

dataset/imagenet.py
AdityaKane2001: Is it okay if I add them a bit later? I am maintaining a doc with all the documentation to add later, but for now I think it would be better to keep the code clean, at least until we know that all of this works well on GCP.

dataset/imagenet.py
AdityaKane2001: I didn't add it right away because the functions have some small bugs. Actually I intend to have all loading and augmentation code related to the ImageNet dataset in this one file. 
Also, a couple of these functions have many redundant cast ops, so I'll clean it up a bit and then add the functions in the `make_dataset` function.

dataset/imagenet.py
sayakpaul: It makes the review a bit harder. So, I would suggest adding the docstrings at this phase.

dataset/imagenet.py
sayakpaul: Okay. Would suggest modularizing whenever possible to keep the length and readability of the scripts accessible.

AdityaKane2001:
@sayakpaul 
Thanks for the quick response!

> @AdityaKane2001 please also add @MorganR as a reviewer.

Morgan has not yet accepted the invitation. I'll add him as soon as he accepts it.

The script for making tfrecords is in [/utils/tfrecords_utils.py](https://github.com/AdityaKane2001/regnety/blob/dev/utils/tfrecords_utils.py)

Sure, I'll use Black.

Sure, I'll add a public Colab notebook for this.

I wanted to ask a couple of things:

1. For testing any of the functions, I do 2 things: Run it through `dataset.map()` and adding a @tf.function, and checking whether the output images are fine. Most of the functions should not be decorated with @tf.function, but I do so just to test whether they work in graph mode. The `read_tfrecs()` function in `imagenet.py` does not play well with @tf.function. Will this be fine in a TPU environment? Does a preprocessing function run in graph mode, even in a TPU system? The dataset gets mapped correctly each time, but as mentioned above, @tf.function gives some issues.
2. What should be the effect of PCA jitter? I processed a couple of images, but they looked unchanged after PCA jitter was applied.

sayakpaul:
> but as mentioned above, @tf.function gives some issues.

What are those issues exactly? 

> Does a preprocessing function run in graph mode, even in a TPU system? 

It does. 

> What should be the effect of PCA jitter? I processed a couple of images, but they looked unchanged after PCA jitter was applied.

Did you run it a couple of times? The jittering effect should be pretty visible. I would suggest comparing the output ranges to the FAIR PCA jitter, which would provide us with some more clues. If you want to get rid of PCA Jitter and instead apply something even more generic and more effective like RandAugment, you can do that. Here's a [blog post](https://keras.io/examples/vision/consistency_training/) that shows how to apply it. We can keep the RandomResized crop function as it is and then apply the RandAugment. 

Also since this is ImageNet-pretraining and using MixUp almost always helps, I would recommend adding a utility for that too. But I totally understand if you'd want to hold off on this.

AdityaKane2001:
It gives the following error:
```
Exception ignored in: <function CapturableResource.__del__ at 0x7f2b77ba2dd0>
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/tensorflow/python/training/tracking/tracking.py", line 269, in __del__
    with self._destruction_context():
  File "/usr/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "/usr/local/lib/python3.7/dist-packages/tensorflow/python/framework/func_graph.py", line 380, in inner_cm
    old_strategy_stack = self._distribution_strategy_stack
  File "/usr/local/lib/python3.7/dist-packages/tensorflow/python/framework/ops.py", line 5195, in _distribution_strategy_stack
    if not hasattr(self._thread_local, "_distribution_strategy_stack"):
AttributeError: 'FuncGraph' object has no attribute '_thread_local'
```
It gives this error when I call the function for the second time.
I think that it gives this error as there's a function inside the main function (code [here](https://github.com/AdityaKane2001/regnety/blob/8c143ca1eacc282342847b9c8bb78a71b4322cbe/dataset/imagenet.py#L41-L59)). I'll remove that and try again.

I agree, it'd be good to apply RandAugment from the start itself. As in the blog post, am I allowed to use the vision library?

I think it would be better to keep Mixup as a backup sort of thing, in case we don't get the required accuracy with the rest. Also, we are training the smaller models, which may give us some room to play with. Training on ImageNette in colab itself will give a good idea, I guess.

AdityaKane2001:
I'm actually new to GCP and would really like some help on that.
I was thinking to make two instances: One with only CPU for TFRecords making and other with the TPU. I'll delete the first instance after we make and store the TFRecords in GCS. Will it be better to do it this way, or to create only one instance and add the TPU later?

dataset/imagenet.py
AdityaKane2001: Okay

sayakpaul: Or even better. Place it under the model itself. Keras now has the following layers:

* [Rescaling](https://www.tensorflow.org/api_docs/python/tf/keras/layers/experimental/preprocessing/Rescaling)
* [Normalization](https://www.tensorflow.org/api_docs/python/tf/keras/layers/experimental/preprocessing/Normalization)

During initializing the `Normalization` layer, we can set the `mean` and `var` arguments to the ImageNet stats. Note that it's variance as opposed to standard deviation (which is typical). So, we would need to square the std.

sayakpaul:
> I agree, it'd be good to apply RandAugment from the start itself. As in the blog post, am I allowed to use the vision library?

Why not? 

One instance should be enough. You can create an `n1-standard-8` instance which is more than enough. You can follow [this guide](https://github.com/GoogleCloudPlatform/training-data-analyst/blob/master/courses/fast-and-lean-data-science/README-TF2.1.md) to see how you can create a TPU-attached Jupyter Notebook instance. Just change the script mentioned in the tutorial to install the latest TF version. Let me know how it goes.

AdityaKane2001:
Okay.

I have received the credits but  I'm not able to redeem them. I have mailed Jaeyoun regarding the same.
I had seen the script earlier, but frankly, I'm more inclined towards executing a script which does the training so that I don't unintentionally overspend.

sayakpaul:
Why would you overspend, though? It's the Compute Engine-based concept with a richer GUI. You could use the Terminal that comes with a Jupyter Lab and run the exact same scripts. It's totally up to you. I find this process easier but I understand if you have other preferences.

AdityaKane2001:
Oh okay, I'll try this.

dataset/imagenet.py
AdityaKane2001: Okay

dataset/imagenet.py
AdityaKane2001: Sure

dataset/imagenet.py
AdityaKane2001: Okay

AdityaKane2001:
@sayakpaul 

>If you want to get rid of PCA Jitter and instead apply something even more generic and more effective like RandAugment, you can do that


As decided, I have removed the PCA jitter function. Over the weekend, I'll
- add RandAugment
- make a notebook for illustration.
- try the pipeline on ImageNette 

I'll request a review again when the pipeline seems good.

/cc @MorganR

AdityaKane2001:
@sayakpaul 
Should I change the directory structure to:

```
regnety/
│   ├── config/ ..
│   ├── dataset/ ..
│   ├── utils/ ..
│   ├── main.py
│   └── (any other files)
├── README.md
├── LICENSE
├── .gitignore
└── (any other files, like tutorials)
```

sayakpaul:
Yeah that looks good to me.

AdityaKane2001:
>I would also like to see a couple of image batches after parsing the TFRecords. A notebook or a demo script might be good to test the data loaders out.


Do we want to create a testing script (like a unittest as you mentioned earlier) or a demo notebook on how to use the repo?

sayakpaul:
A unit test sounds good. Notebook is also good to just interact and debug things.

AdityaKane2001:
Ok, then I'll add unit tests for now, will add the illustrative notebook later.

sayakpaul:
Okay. But for now, I think it would be better if you just create a rough notebook to ensure the dataloader is working as expected. This will save you a lot of time during model training if something goes wrong because of data preprocessing.

AdityaKane2001:
Oh, okay sure.

AdityaKane2001:
@sayakpaul @MorganR  
Today I was planning to fire up the instance and make the TFRecords. Could you please tell if [this](https://github.com/AdityaKane2001/regnety/blob/dev/regnety/utils/tfrecords_utils.py) script is good to go?

sayakpaul:
> @sayakpaul @MorganR
> Today I was planning to fire up the instance and make the TFRecords. Could you please tell if [this](https://github.com/AdityaKane2001/regnety/blob/dev/regnety/utils/tfrecords_utils.py) script is good to go?

Looks good to me but here are a couple of things you could do:

* Don't shuffle the indices for non-train datasets. 
* Prefer having a batch of entries inside a single TFRecord. You can create a `tf.data.Dataset` of the image paths first, batch it, and then _map_ your TFRecord utils to it. [Here's](https://github.com/sayakpaul/TF-2.0-Hacks/blob/master/Cats_vs_Dogs_TFRecords.ipynb) an example. This not only will improve your TFRecord creation process but will also speed up the data input pipeline. 

Let me know if anything unclear.

AdityaKane2001:
Got it, thanks!

AdityaKane2001:
@sayakpaul 
>You can create a tf.data.Dataset of the image paths first, batch it, and then map your TFRecord utils to it


Please take a look at the following code:

```python
def make_image(filepath):

    image_str = tf.io.read_file(filepath)

    if is_png(filepath):
        image_str = png_to_jpeg(image_str)

    if is_cmyk(filepath):
        image_str = cmyk_to_rgb(image_str)


    image_tensor = tf.io.decode_jpeg(image_str)

    height = image_tensor.get_shape()[0] #Here's the issue
    width = image_tensor.get_shape()[1]

    if image_tensor.shape[2] == 1:
        image_tensor = tf.image.grayscale_to_rgb(image_tensor)
        image_str = tf.io.encode_jpeg(image_tensor)

    assert len(image_tensor.shape) == 3

    return image_str, height, width

```

Since `image_tensor` is defined during execution itself, it's shape is not predefined. Thus, when I try to map this function to the filepaths dataset, height and width come out to be `None`. Wrapping it with `tf.py_function` will not help with speed, as it will execute it eagerly.  Could you please shed some light on this?

sayakpaul:
What's preventing us from resizing it? When using TPUs, the shapes must be explicitly specified anyway.

AdityaKane2001:
We can do that. I'll  convert all images to 512x512, as most images are 375x500 or similar. 
I'll use the filter to check whether the image is grayscale and convert it to RGB.
Thanks!

AdityaKane2001:
@sayakpaul 
Could you please take a look again? I have made changes in [tfrecords_utils.py](https://github.com/AdityaKane2001/regnety/blob/dev/regnety/utils/tfrecords_utils.py) as discussed.

sayakpaul:
Thanks for making the changes. I think @MorganR will have additional comments. So I will for them.

MorganR:
Looking now

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,268 @@
...
+        tfrecs_filepath: list of filepaths of all TFRecords files
+        batch_size: batch_size for the Dataset
+        image_size: final image size of the images in the dataset
+        augment_fn: function to apply to dataset after loading raw TFrecords
+                    'default' implies following augmentations will be applied:
+                    - Random sized crop (train only)
+                    - Scale and center crop
+                    - RandAugment (TODO)
+        num_classes: number of classes
+        randaugment: True of RandAugment is to be applied
```
MorganR, line 21: nit: if

AdityaKane2001: Made requested changes

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,268 @@
...
+        return {
+            "image": image,
+            "height": self.image_size,
+            "width": self.image_size,
+            "filename": example["filename"],
+            "label": example["label"],
+            "synset": example["synset"],
+        }
+
+    def clip_example(self, example):
```
sayakpaul, line 214: I don't it's a valid name for a function being used to one hot encode labels.

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,268 @@
...
+        if tfrecs_filepath is None:
+            raise ValueError("List of TFrecords paths cannot be None")
+        self.tfrecs_filepath = tfrecs_filepath
+        self.batch_size = batch_size
+        self.image_size = image_size
+        self.augment_fn = augment_fn
+        self.num_classes = num_classes
+        self.randaugment = randaugment
+        if self.randaugment:
+            self.augmenter = RandAugment(magnitude=5)
```
sayakpaul, line 61: It would also make sense to specify the `layers` argument here to denote the width of the augmentation chain.

regnety/main.py
```diff
@@ -0,0 +1,42 @@
...
+        
+
+    imgnet = ImageNet([os.path.join('/content',i) for i in os.listdir('/content') if i.startswith('trial5')])
+    ds = imgnet.make_dataset()
+    for i in ds:
+      print(i)
+      im = i['image']/255.
+      plt.imshow(im)
+      plt.savefig('image.jpeg')
+      break
```
sayakpaul, line 39: After we have verified the results make sure this block is ommitted.

regnety/utils/tfrecords_utils.py
```diff
@@ -0,0 +1,184 @@
...
+    
+    num_shards = int(math.ceil(len(images) / batch_size))
+
+    shard = 0
+    for (image_str, filename, label, synset) in ds:
+        
+        output_filepath = os.path.join(
+            output_dir, file_prefix + "_%.4d_of_%.4d.tfrecord" % (shard, num_shards)
+        )
+        print("Writing %d of %d shards"%(shard, num_shards))
```
sayakpaul, line 174: Given the size of ImageNet-1k the logs can easily get messy. Do we want to set a frequency interval here? Also, do we have a driver script for this utility?

sayakpaul (changes requested):

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,268 @@
...
+            A tf.data.Dataset
+        """
+        ds = tf.data.TFRecordDataset(self.tfrecs_filepath)
+        ds = ds.map(
+            lambda example: tf.io.parse_example(example, ImageNet.TFRECS_FORMAT)
+        )
+        ds = ds.map(lambda example: self.decode_example(example))
+        return ds
+
+    def _scale_and_center_crop(self, image, h, w, scale_size, final_size):
```
MorganR, line 101: A downside of this approach is you are always cropping non-square images. What do you think of setting the *longer* side to `scale_size`, and adding black padding to the shorter side? You could remove most of your logic and achieve this by simply calling:

```python
square_scaled_image = tf.image.resize_with_pad(image, scale_size, scale_size)
return tf.image.central_crop(tf.cast(final_size, tf.float32) / scale_size)
```

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,268 @@
...
+            im = tf.image.resize(image, (w, h))
+        else:
+            im = tf.image.resize(image, (w, h))
+        x = tf.cast((w - final_size) / 2, tf.int64)
+        y = tf.cast((h - final_size) / 2, tf.int64)
+        return im[y : (y + final_size), x : (x + final_size), :]
+
+    def random_sized_crop(self, example, min_area=0.08, max_iter=10):
+        """
+        Randomly chooses an area and aspect ratio and resizes the image to those
```
MorganR, line 133: aspect ratio is always square, right?

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,268 @@
...
+        values.
+
+        Args:
+            example: A dataset example.
+            min_area: Minimum area of image to be used
+            max_iter: Maximum iteration to try before implementing scale and
+                center crop. Randomly genereted width and height must be lower
```
MorganR, line 140: nit: generated

regnety/dataset/imagenet.py
AdityaKane2001: I'll change it to `make_one_hot_targets`

regnety/dataset/imagenet.py
AdityaKane2001: I'll add that

regnety/main.py
AdityaKane2001: Yes

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,268 @@
...
+        """
+
+        h, w = tf.cast(example["height"], tf.int64), tf.cast(
+            example["width"], tf.int64
+        )
+        area = h * w
+        return_example = False
+        image = example["image"]
+        num_iter = tf.constant(0, dtype=tf.int32)
+        while num_iter <= max_iter:
```
MorganR, line 154: Looping like this for randomization is a little wasteful and error-prone, especially since this will run many times. It's always best to avoid this if you can. Also note that our final image is always square, therefore we only need to choose one "size" to extract.

Consider a simplified approach, like:

```python
min_crop_size = math.sqrt(min_area * h * w)
# assuming you allow padding:
max_crop_size = max(h, w)
crop_size = tf.random.uniform((), minval=min_crop_size, maxval=max_crop_size, dtype=tf.int64)
# Note: you might need to pad or crop the image to make it square before calling random_crop.
crop = tf.image.random_crop(image, size=(crop_size, crop_size, 3))

# Now resize if necessary (assuming this function should always return images with a constant size)
```

regnety/utils/tfrecords_utils.py
AdityaKane2001: I'll add a frequency argument
No, we don't. I'll add a shell script in `utils` folder. The script will be for
- downloading ImageNet1k
- making TFRecords
- creating GCS bucket
- copying TFRecords to the bucket
- shutting down the VM

Is that okay?

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,268 @@
...
+        return {
+            "image": image,
+            "height": self.image_size,
+            "width": self.image_size,
+            "filename": example["filename"],
+            "label": example["label"],
+            "synset": example["synset"],
+        }
+
+    def clip_example(self, example):
```
MorganR, line 214: Why is this called "clip_example"? The name doesn't make sense to me. Maybe "one_hot_encode_example"?

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,268 @@
...
+        if tfrecs_filepath is None:
+            raise ValueError("List of TFrecords paths cannot be None")
+        self.tfrecs_filepath = tfrecs_filepath
+        self.batch_size = batch_size
+        self.image_size = image_size
+        self.augment_fn = augment_fn
+        self.num_classes = num_classes
+        self.randaugment = randaugment
+        if self.randaugment:
+            self.augmenter = RandAugment(magnitude=5)
```
MorganR, line 61: nit: augmenter should be private (self._augmenter)

regnety/utils/tfrecords_utils.py
sayakpaul: I would prefer the following to be done manually:

* downloading ImageNet1k
* creating GCS bucket

This way the end-users will have a bit more flexibility in terms of options I believe. @MorganR would do you think?

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,268 @@
...
+                'target' : (num_classes,)
+            }
+        """
+        ds = self._read_tfrecs()
+
+        if self.augment_fn == "default":
+            ds = ds.map(lambda example: self.random_sized_crop(example))
+            if self.randaugment:
+                ds = ds.map(lambda example: self._randaugment(example))
+            #ds = ds.map(lambda example: self.clip_example(example))
```
MorganR, line 263: Should this be deleted? Also consider deleting the `self.clip_example` method if it's unused

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,268 @@
...
+
+        else:
+            ds = ds.map(lambda example: self.augment_fn(example))
```
MorganR, line 266: It seems like randaugment should still be applied if self.randaugment is true. Otherwise it is strange for that argument to be provided at all.

regnety/utils/tfrecords_utils.py
AdityaKane2001: Okay

regnety/main.py
```diff
@@ -0,0 +1,42 @@
...
+
+    args = parser.parse_args()
+    make_tfrecs(
+        dataset_base_dir=args.data_dir,
+        output_dir=args.odir,
+        file_prefix=args.file_prefix,
+        synset_filepath=args.synset_filepath,
+        batch_size = args.batch_size,
+    )
+    # for i in ds:
```
MorganR, line 27: Please delete

regnety/utils/tfrecords_utils.py
```diff
@@ -0,0 +1,184 @@
...
+    return tf.train.Feature(float_list=tf.train.FloatList(value=[value]))
+
+
+def _int64_feature(value):
+    """Returns an int64_list from a bool / enum / int / uint."""
+    return tf.train.Feature(int64_list=tf.train.Int64List(value=[value]))
+
+
+# make_tfrecs -> for(make_single_tfrecord)-> make_dataset -> make_example
+def get_synset_labels(filepath):
```
MorganR, line 32: Can this be private?

regnety/utils/tfrecords_utils.py
```diff
@@ -0,0 +1,184 @@
...
+
+    for i in raw_labels_dict:
+        labels_dict[raw_labels_dict[i]["id"]] = (
+            int(i),
+            raw_labels_dict[i]["label"],
+        )
+    return labels_dict
+
+
+def get_files(data_dir, synset_filepath, shuffle = True):
```
MorganR, line 46: Can this be private?

regnety/utils/tfrecords_utils.py
```diff
@@ -0,0 +1,184 @@
...
+    all_images = tf.io.gfile.glob(os.path.join(data_dir, "*", "*.JPEG"))
+    all_synsets = [os.path.basename(os.path.dirname(f)) for f in all_images]
+
+    print(len(all_images))
```
MorganR, line 50: Should the print still be here? If so, consider adding some more context so it's clear what the number means.

regnety/utils/tfrecords_utils.py
```diff
@@ -0,0 +1,184 @@
...
+    all_synsets = [all_synsets[i][1:] + "-n" for i in all_indexes]
+
+    labels_dict = get_synset_labels(synset_filepath)
+
+    all_labels_int = [labels_dict[i][0] for i in all_synsets]
+
+    return all_images, all_labels_int, all_synsets
+
+
+def make_image(filepath):
```
MorganR, line 65: Private?

regnety/utils/tfrecords_utils.py
```diff
@@ -0,0 +1,184 @@
...
+    image_tensor = tf.cast(tf.image.resize(image_tensor, (512,512)), tf.uint8)
+
+    image_str = tf.io.encode_jpeg(image_tensor)
+
+    assert len(image_tensor.shape) == 3
+
+    return image_str
+
+
+def make_example(image_str, filepath, label, synset):
```
MorganR, line 90: Private?

regnety/utils/tfrecords_utils.py
```diff
@@ -0,0 +1,184 @@
...
+                    ),
+                    "label": _int64_feature(label),
+                    "synset": _bytes_feature(synset),
+                }
+            )
+        )
+    return example
+
+
+def make_single_tfrecord(
```
MorganR, line 121: Private?

regnety/utils/tfrecords_utils.py
```diff
@@ -0,0 +1,184 @@
...
+
+    if is_cmyk(filepath):
+        image_str = cmyk_to_rgb(image_str)
+
+    image_tensor = tf.io.decode_jpeg(image_str)
+
+    if not is_rgb(image_tensor):
+        image_tensor = tf.image.grayscale_to_rgb(image_tensor)
+        
+    image_tensor = tf.cast(tf.image.resize(image_tensor, (512,512)), tf.uint8)
```
MorganR, line 81: Do you think the size here should be configurable? Also if these images aren't square, this method breaks the aspect ratio. Consider using resize_with_pad to preserve the images.

regnety/utils/tfrecords_utils.py
```diff
@@ -0,0 +1,184 @@
...
+                    "image": _bytes_feature(image_str),
+                    "filename": _bytes_feature(
+                        os.path.basename(filepath)
+                    ),
+                    "label": _int64_feature(label),
+                    "synset": _bytes_feature(synset),
+                }
+            )
+        )
+    except:
```
MorganR, line 105: Why is this try except block here? The code is exactly the same in each. What problem does this solve?

regnety/utils/tfrecords_utils.py
```diff
@@ -0,0 +1,184 @@
...
+
+    filepaths_ds = tf.data.Dataset.from_tensor_slices([
+        (images[i]) for i in range(len(labels))
+    ])
+
+
+    images_ds = filepaths_ds.map(lambda filepath: make_image(filepath))
+    labels_ds = tf.data.Dataset.from_tensor_slices(labels)
+    synsets_ds = tf.data.Dataset.from_tensor_slices(synsets)
+    filepaths_ds = filepaths_ds.map(lambda filepath: tf.strings.split(filepath, '/')[-1])
```
MorganR, line 161: nit: `filename_ds` would be more clear

regnety/utils/tfrecords_utils.py
MorganR, line 174: Please be consistent with spacing (eg. either put spaces around the `%` sign as above, or remove them above).

regnety/utils/tfrecords_utils.py
```diff
@@ -0,0 +1,184 @@
...
+
+            example = make_example(
+                image_str, height, width, chunk_files[i], label, synset
+            )
+
+            writer.write(example.SerializeToString())
+    writer.close()
+
+
+def make_tfrecs(
```
MorganR, line 139: Please add a docstring to this function

regnety/main.py
```diff
@@ -0,0 +1,42 @@
...
+    #     print(i)
+    #     break
+        
+
+    imgnet = ImageNet([os.path.join('/content',i) for i in os.listdir('/content') if i.startswith('trial5')])
```
MorganR, line 32: I'm a bit confused about the point of this script. If I understand correctly:

`make_tfrecs` reads the ImageNet data, preprocesses it to a consistent format and image size, and saves that as TFRecords. That seems to be all that's needed for one-time data preprocessing.

Then this code runs to augment the images. Why is that happening at the same time? As discussed in our last meeting, it doesn't make sense to perform augmentations such as random cropping once during pre-processing. These augmentations should happen continuously during training.

MorganR: To elaborate a little more, performing random augmentation now just goes from image set A with size `n` to slightly different image set B also with size `n`. This has minimal benefit. Conversely, performing random augmentations continuously during training means that in each epoch you go from image set A of size `n` to some new image set B with size `n`, and then on the next epoch you get new image set C, then D, etc. This effectively multiplies the number and variety of training examples you have.

MorganR (changes requested):

regnety/utils/tfrecords_utils.py
MorganR: +1, doing those manually is simpler

regnety/main.py
AdityaKane2001: `main.py` is a script that I just use to test all the functions quickly and as a pipeline. Yes, only the preprocessing part (i.e. converting raw data to TFRecords) will be done beforehand. The `ImageNet` class is for training time.

regnety/utils/tfrecords_utils.py
AdityaKane2001: I'll add docstrings to all functions in `tfrecords_utils.py` today itself.

regnety/utils/tfrecords_utils.py
AdityaKane2001: Yes

regnety/utils/tfrecords_utils.py
AdityaKane2001: Okay

regnety/utils/tfrecords_utils.py
AdityaKane2001: I'll remove it. It was needed when the script was meant to be ran in Eager mode.

regnety/utils/tfrecords_utils.py
AdityaKane2001: Okay, will do those things manually. For the following things, I'll make a shell script
- making TFRecords
- copying TFRecords to the bucket
- shutting down the VM

regnety/utils/tfrecords_utils.py
AdityaKane2001: Sure

regnety/utils/tfrecords_utils.py
AdityaKane2001: Sure

regnety/utils/tfrecords_utils.py
AdityaKane2001: No, it was for debugging. I'll remove it.

regnety/utils/tfrecords_utils.py
AdityaKane2001: Yes

regnety/utils/tfrecords_utils.py
AdityaKane2001: Yes

regnety/dataset/imagenet.py
AdityaKane2001: I'll change it

regnety/dataset/imagenet.py
AdityaKane2001: Yes, the approach seemed off to me, too, but I decided to stick with it as it was in the official repo. 

As far as I understand, the objective of this function is to distort the scale of the images (i.e. to stretch or squeeze the images) and then take a random crop. That's why I avoided smart_resize or any other method which preserves the scale. Please correct me if I am wrong.

regnety/dataset/imagenet.py
AdityaKane2001: Noted

regnety/dataset/imagenet.py
AdityaKane2001: Aspect ratio is a loose term. I mean to say the attribute of the image that it's stretched or squeezed.

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,268 @@
...
+    def _scale_and_center_crop(self, image, h, w, scale_size, final_size):
+        """Resizes image to given scale size and returns a center crop. Aspect
+        ratio is NOT maintained.
+
+        Args:
+            image: tensor of the image
+            h: initial height of image
+            w: initial width of image
+            scale_size: Size of image to scale to
+            final_size: Size of final image
```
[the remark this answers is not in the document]
AdityaKane2001: I'll add that

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,268 @@
...
+        Returns:
+            Tensor of shape (final_size, final_size, 3)
+        """
+        if w < h and w != scale_size:
+            w, h = tf.cast(scale_size, tf.int64), tf.cast(
+                ((h / w) * scale_size), tf.int64
+            )
+            im = tf.image.resize(image, (w, h))
+        elif h <= w and h != scale_size:
+            w, h = tf.cast(((h / w) * scale_size), tf.int64), tf.cast(
```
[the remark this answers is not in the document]
AdityaKane2001: I'll change that

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,268 @@
...
+        ds = tf.data.TFRecordDataset(self.tfrecs_filepath)
+        ds = ds.map(
+            lambda example: tf.io.parse_example(example, ImageNet.TFRECS_FORMAT)
+        )
+        ds = ds.map(lambda example: self.decode_example(example))
+        return ds
+
+    def _scale_and_center_crop(self, image, h, w, scale_size, final_size):
+        """Resizes image to given scale size and returns a center crop. Aspect
+        ratio is NOT maintained.
```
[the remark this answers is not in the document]
AdityaKane2001: Sorry about that. Messed up `_scale_and_center_crop` and `random_sized_crop`. Yes, aspect ratio is maintained in this function.

regnety/dataset/imagenet.py
AdityaKane2001: Sure, that sounds good

regnety/dataset/imagenet.py
AdityaKane2001: Okay

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,268 @@
...
+        "image": tf.io.FixedLenFeature([], tf.string),
+        # "height": tf.io.FixedLenFeature([], tf.int64),
+        # "width": tf.io.FixedLenFeature([], tf.int64),
+        "filename": tf.io.FixedLenFeature([], tf.string),
+        "label": tf.io.FixedLenFeature([], tf.int64),
+        "synset": tf.io.FixedLenFeature([], tf.string),
+    }
+
+    _MEAN = tf.constant([0.485, 0.456, 0.406])
+    _STD = tf.constant([0.229, 0.224, 0.225])
```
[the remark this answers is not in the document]
AdityaKane2001: Okay, I'll change that

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def decode_example(self, example):
+        """Decodes an example to its individual attributes
+
+        Args:
+            example: A TFRecord dataset example.
+
+        Returns:
+            Dict containing attributes from a single example. Follows
+            the same names as TFRECORDS_FORMAT.
```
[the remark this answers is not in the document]
AdityaKane2001: > It looks like this return data isn't a perfect match for what's inside the TFRecord

I changed the return data format slightly because it was decided to resize the images to 512. Thus, height and width become implicit (for our use case). This is in continuation to line 81 in `tfrecords_utils.py` .

regnety/utils/tfrecords_utils.py
AdityaKane2001: Please take a look at [this](https://github.com/AdityaKane2001/regnety/pull/1/files/40f93df36d97c005ceb0eb8d8a0f6c66878dede4#r650963897).

AdityaKane2001:
@sayakpaul  @MorganR 

I wanted to discuss about resizing of the images before saving them to TFRecords. Personally, I think that we should keep the stock images as they are, even at the cost of performance during TFRecords creation, since it is one-time. Resizing the images may result in worse augmentation, as the images will be distorted before `RandAugment` and `random scale and crop`. 

Also, a small doubt, as far as I understand the objective of [random_sized_crop](https://github.com/AdityaKane2001/regnety/blob/d618531f6c4a758d403d8572f4510ea6c810243f/regnety/dataset/imagenet.py#L131-L213) function is to stretch or squeeze the image and then take a random square crop. Is this correct?

Please share your thoughts on the same and please correct me if I am wrong.

sayakpaul:
> I wanted to discuss about resizing of the images before saving them to TFRecords. Personally, I think that we should keep the stock images as they are, even at the cost of performance during TFRecords creation, since it is one-time. Resizing the images may result in worse augmentation, as the images will be distorted before RandAugment and random scale and crop.

I don't you will be able to batch together images of different sizes and serialize those as a single TFRecord shard. In my experience and also from what I have mostly seen this resizing is not that hurtful. Here's a [reference](https://github.com/tensorflow/tpu/blob/master/models/official/densenet/densenet_imagenet.py). Feel free to also refer to how it's done in `models/official`. @MorganR feel free to chime in here. 

> Also, a small doubt, as far as I understand the objective of random_sized_crop function is to take a random square crop and then stretch or squeeze the image. Is this correct?

Yes. It should ideally replicate something like [this](https://github.com/tensorflow/tpu/blob/master/models/official/densenet/vgg_preprocessing.py#L45).

AdityaKane2001:
@MorganR 
I was following [this](https://github.com/facebookresearch/pycls/blob/f20820e01eef7b9a47b77f13464e3e77c44d5e1f/pycls/datasets/transforms.py#L32-L48) code.

regnety/dataset/imagenet.py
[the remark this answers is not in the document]
AdityaKane2001: >consider defining a simple dataclass specifically for this return value.

@MorganR 
If I understand correctly, you were saying that I should make a dataclass and use it as a common return object for all functions which we map to the dataset (like `decode_example` in this case). In that case, I think tf.data.Dataset allows only tuples,lists or dicts as its elements. Please correct me if I'm wrong.

>This would provide an even better IDE experience if you add type annotations


Sure, I'll add type annotations.

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,268 @@
...
+        """
+        image = tf.cast(tf.io.decode_jpeg(example["image"]), tf.float32)
+        height = 512
```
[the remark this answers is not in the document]
AdityaKane2001: Yes

AdityaKane2001:
@sayakpaul @MorganR 
Please take a look at the updated code. Following are the important changes:

- `dataset.imagenet.ImageNet.random_sized_crop` uses `tf.image.sample_distorted_bounding_box`  , which I think was the objective of [this](https://github.com/facebookresearch/pycls/blob/f20820e01eef7b9a47b77f13464e3e77c44d5e1f/pycls/datasets/transforms.py#L32-L48). Code is based on tensorflow/models' [implementation ](https://github.com/tensorflow/models/blob/e39f2477ac207b2d0e15a8d9f172e07e76799eb0/research/slim/preprocessing/inception_preprocessing.py#L99-L255) of the same.
- `dataset.imagenet.ImageNet.scale_and_center_crop`  is now simpler.
- Added a separate python script to be used for making TFRecords. ([/regnety/make_tfrecords.py](https://github.com/AdityaKane2001/regnety/blob/dev/regnety/make_tfrecords.py))
- Added type annotations.
- Made a notebook for testing the dataloader, gist [here](https://colab.research.google.com/gist/AdityaKane2001/3d40273d74143d1488529062c6851937/regnety_dataloader.ipynb?authuser=1).

I also have a couple of things to discuss. In previous versions, there was an argument for synset json file path, even though we have that json file in config folder. I have used a small [hack](https://github.com/AdityaKane2001/regnety/blob/dbc35730934690a1b42df1dcd92d4097beca1b83/regnety/utils/tfrecords_utils.py#L56-L60) to get this path for now without any external input. Going forward, what would be the best way to handle configurations? Also, will it be better if I make a setup.py file and use regnety as a module?

sayakpaul:
> I also have a couple of things to discuss. In previous versions, there was an argument for synset json file path, even though we have that json file in config folder. I have used a small hack to get this path for now without any external input. Going forward, what would be the best way to handle configurations?

For data-related stuff, using separate classes decorated with [`dataclasses`](https://docs.python.org/3/library/dataclasses.html) would be a good idea. In general, for handling configuration-related stuff I would recommend either ` config.py` or `config.yaml`. 

> Also, will it be better if I make a setup.py file and use regnety as a module?

Could you provide some pseudocode in Python to help us understand this better?

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,223 @@
...
+}
+    
+
+class ImageNet:
+    """Class for all ImageNet related functions, includes TFRecords loading,
+    preprocessing and augmentations. TFRecords must follow the format given
+    below. If not specified otherwise in `augment_fn` argument, following 
+    augmentations are applied to the dataset:
+    - Random sized crop (train only)
+    - Scale and center crop
```
sayakpaul, line 24: Are both of these applied? Could you be a bit more explicit about this?

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,223 @@
...
+        "width": tf.io.FixedLenFeature([], tf.int64),
+        "filename": tf.io.FixedLenFeature([], tf.string),
+        "label": tf.io.FixedLenFeature([], tf.int64),
+        "synset": tf.io.FixedLenFeature([], tf.string),
+}
+    
+
+class ImageNet:
+    """Class for all ImageNet related functions, includes TFRecords loading,
+    preprocessing and augmentations. TFRecords must follow the format given
```
sayakpaul, line 20: How about changing the first sentence to the following?

```
Class for all ImageNet data-related functions, including TFRecord parsing along with augmentation transforms. 
```

Where have you specified the TFRecord format?

regnety/dataset/imagenet.py
```diff
@@ -0,0 +1,223 @@
...
+    def _read_tfrecs(self) -> Type[tf.data.Dataset]:
+        """Function for reading and loading TFRecords into a tf.data.Dataset.
+
+        Returns:
+            A tf.data.Dataset
+        """
+        ds = tf.data.TFRecordDataset(self.tfrecs_filepath)
+        ds = ds.map(
+            lambda example: tf.io.parse_example(example, _TFRECS_FORMAT)
+        )
```
sayakpaul, line 92: Parallelize this `map()` call by specifying `num_parallel_calls=tf.data.AUTOTUNE`. [Here's](https://www.tensorflow.org/guide/data_performance) the official guide on `tf.data` to get the most out of it in terms of performance.

regnety/main.py
```diff
@@ -0,0 +1,63 @@
...
+import shutil
+import random
+import json
+import argparse
+import math
+from utils.tfrecords_utils import *
+from dataset.imagenet import ImageNet
+import matplotlib.pyplot as plt
+import time
+from datetime import timedelta
```
sayakpaul, line 12: Organize the imports.

regnety/utils/tfrecords_utils.py
```diff
@@ -0,0 +1,288 @@
...
+            image_str, height, width = _make_image(chunk_files[i])
+            label = chunk_labels[i]
+            synset = chunk_synsets[i]
+
+            example = _make_example(
+                image_str, height, width, chunk_files[i], label, synset
+            )
+
+            writer.write(example.SerializeToString())
+    writer.close()
```
sayakpaul, line 220: Should we print how many TFRecords have been created within some prespecified interval?

sayakpaul (changes requested):

regnety/dataset/imagenet.py
AdityaKane2001: I'll change that to
```
- Random sized crop (train only)
- Scale and center crop (validation and test only)
```

regnety/dataset/imagenet.py
AdityaKane2001: Sure, I'll change that

regnety/dataset/imagenet.py
AdityaKane2001: Okay

regnety/main.py
AdityaKane2001: Okay

regnety/utils/tfrecords_utils.py
AdityaKane2001: Actually, I have added the `--log_freq` flag. I don't think it is required for now, maybe I can add it later on if it feels necessary. Is that okay?

sayakpaul: How are you planning to keep track of the progress then?

AdityaKane2001: That didn't cross my mind, sorry about that. I'll add this.

AdityaKane2001:
>  Also, will it be better if I make a setup.py file and use regnety as a module?

Following is our file structure now. I was thinking of adding a new file `setup.py` as follows:
```
regnety
├── regnety/
│   ├── config/
│   │   └── imagenet_synset_to_human.json
│   ├── dataset/
│   │   ├── __init__.py
│   │   └── imagenet.py
│   ├── utils/
│   │   ├── __init__.py
│   │   ├── image_utils.py
│   │   └── tfrecord_utils.py
│   ├── main.py
│   └── make_tfrecords.py
├── setup.py ###
├── README.md
├── LICENSE
└── .gitignore
```
where setup.py has
```python
from setuptools import setup

setup(
    name = 'regnety',
    package_dir = {'' : 'regnety'},
    packages = ['utils', 'dataset']
)
```

By doing so, I'll be able to use, say for example 
```python
#/regnety/dataset/imagenet.py

from utils import image_utils
...
```
The other way is to just append the absolute regnety path in PATH in `regnety.__init__.py`, but I thought this might be simpler.


>Try suppressing all the TensorFlow warnings that harmless. Give a descriptive message beside 9469. Append a unit beside the total time taken to prepare the shards.

I'll add these things along with the other changes in the next commit.

regnety/dataset/imagenet.py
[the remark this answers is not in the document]
MorganR: Makes sense. I added this comment before reading the rest of the files.

regnety/utils/tfrecords_utils.py
AdityaKane2001: > Should we print how many TFRecords have been created within some prespecified interval?

How can we do this elegantly? I thought of two naive solutions, but they both have their issues:

1. Check whether  the predefined interval time has passed sice the last execution of `_make_example` function (it's the function that takes the least time. ) Downside: too many arguments.
2. Same as 1, but use global variables instead of arguments. 

I looked at [sched](https://docs.python.org/3/library/sched.html) and [schedule](https://github.com/dbader/schedule) libraries. None of those provide the ability to pass new arguments on each execution.
Neither of the above are good approaches. Could you please suggest a solution?

sayakpaul: If the user asks for logging the frequency we log it with respect to an interval we specify. Not letting the user specify this interval is okay IMO. 

@MorganR any particular improvement you'd suggest here?

AdityaKane2001: I don't think that the interval being defined by the user or not is the issue. The issue is that when we implement this (using one of the methods mentioned in my earlier comment), the code is a bit less readable. I didn't find any concrete workaround for that. Here's the old [file](https://github.com/AdityaKane2001/regnety/blob/dev/regnety/utils/tfrecords_utils.py) and [here's](https://gist.github.com/AdityaKane2001/ba11ff7f34a8da86e8cb511bc9d03942) the one with the changes. I'll commit these changes if this seems alright.

sayakpaul: What's stopping us from doing something like so?

```python
if iteration % freq_interval == 0:
    print(f"Serialized {count} TFRecords."
```

`count` could be a global variable here. To my eyes, the code is not anyway unreadable. You can find a complete example here in [this notebook](https://nbviewer.jupyter.org/github/sayakpaul/TF-2.0-Hacks/blob/master/Cats_vs_Dogs_TFRecords.ipynb).

AdityaKane2001: I have already implemented this. Code [here](https://github.com/AdityaKane2001/regnety/blob/3d8b1e4636611b205c2b9f09946e18eed311d4f2/regnety/utils/tfrecords_utils.py#L278-L280). I could change this to print after the TFRecord is made.

sayakpaul: Perfect then.

AdityaKane2001: Great, thanks! I'll resolve this.

regnety/utils/image_utils.py
```diff
@@ -13,7 +13,8 @@ def is_png(filename):
 
 
 def is_cmyk(filename):
-    blacklist = tf.constant([
+    blacklist = tf.constant(
```
MorganR, line 16: Consider a more specific term than `blacklist`, like: `cmyk_filenames`. It's more clear, and avoids potential offense.

As a general note, this kind of change is happening across the industry and is good to become aware of. Some examples:

 * [A recent Rails issue](https://github.com/rails/rails/issues/33677)
 * [Some Chromium changes](https://www.theregister.com/2019/09/03/chromium_microsoft_offensive/)

MorganR (approved):
Looks much improved, thanks Aditya! Just a few minor suggestions

regnety/dataset/imagenet.py
```diff
@@ -257,12 +213,26 @@ def make_dataset(
...
+                num_parallel_calls = tf.data.AUTOTUNE
+            )
+
+        if self.randaugment:
+            ds = ds.map(
+                lambda example: self._randaugment(example),
+                num_parallel_calls = tf.data.AUTOTUNE
+            )
+        
+        # ds = ds.map(
```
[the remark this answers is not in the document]
AdityaKane2001: No, I have kept it that way for illustration right now. I'll uncomment it before actually executing the script.

regnety/utils/image_utils.py
AdityaKane2001: Sorry about that. I'll change it right away.

regnety/dataset/imagenet.py
AdityaKane2001: @sayakpaul 
AFAIK the map function records all the calls and employs a model to optimize it. This may be a silly question, but do we parallelize every function or some selective ones?

sayakpaul:
@AdityaKane2001 since we are almost about to merge this PR, please consider submitting smaller PRs in the coming days. That way it's easier and more compact for us to review and work on.

AdityaKane2001:
Sure, I'll keep one file at a time.

sayakpaul:
Also, if you agree with a particular suggestion, you can just thumb it up. That way the conversation length is maintained and things look cleaner. Of course, if you think a suggestion needs more clarification or you need to brainstorm more, comment without any hesitation.

sayakpaul (approved):

regnety/dataset/imagenet.py
sayakpaul: > AFAIK the map function records all the calls and employs a model to optimize it. 

Could you expand a bit more on this with some credible sources? 

> This may be a silly question, but do we parallelize every function or some selective ones?

Not silly at all. If we mindlessly parallelize things then it may yield a negative effect. That said, in this case parallelizing the record parsing call won't be harmful.

AdityaKane2001:
Thanks a lot for the reviews! I'll merge the PR now. 
@sayakpaul  [Here's](https://colab.research.google.com/gist/AdityaKane2001/4b4e32d45cb95f46bba59e43e35c63f1/regnety_dataloader.ipynb?authuser=1) the demo notebook.

regnety/dataset/imagenet.py
AdityaKane2001: Sorry for delay in the reply, I didn't see the comment earlier.

>Could you expand a bit more on this with some credible sources?

"_The auto-tuning implementation uses the processing time
and the input pipeline structure to build an analytical model
that is used to estimate how input pipeline parameters affect
end-to-end latency_" - Read this from [here](https://arxiv.org/pdf/2101.12127.pdf).

> If we mindlessly parallelize things then it may yield a negative effect.

Could you please check the code once again? I have added `num_parallel_calls` almost everywhere. Also, I just realized,  is `random_sized_crop`  and `scale_and_center`  traceable functions? I mean, `example` is a dict which is a part of the tf.data.dataset but its individual image tensors may vary in size. So will the function execute without making too many graphs (and possibly OOM)?

sayakpaul: But we don't actually ask it to auto-tune it, it's won't, right? That is the entire point of using `AUTOTUNE`. 

> Could you please check the code once again? I have added num_parallel_calls almost everywhere. Also, I just realized, is random_sized_crop and scale_and_center traceable functions? I mean, example is a dict which is a part of the tf.data.dataset but its individual image tensors may vary in size. So will the function execute without making too many graphs (and possibly OOM)?

To confirm this you can debug the input bottleneck inside TensorBoard. Here's the guide: https://www.tensorflow.org/guide/data_performance.

Issue #2, opened 2021-06-21
Upcoming milestones


AdityaKane2001:
Hello @sayakpaul and @MorganR ,

Following are the milestones for the first couple of weeks.  

Data preprocessing and input pipeline implementation and testing (Weeks 1 and 2)
1.  **Augmentations to be applied** 
2. **TFRecords config**
3. **Testing speed and correctness of input pipeline**
4. GCS bucket config
5. Configuration of GCP VM and TPU 

I believe the first three of them are more or less completed, and we are now ironing out the finer details. They will be formally completed when the PR is merged. The latter two will be completed as soon as I have the GCP access, and should ideally be done within three days at the most. 
This brings us to the next couple of weeks and the milestones are:

Actual model scripting and testing (Weeks 3 and 4)
1. Implementation of base level classes (Block, Stage, etc.) 
2. Implementation of 4 RegNetY variations
3. Testing the model(s) on small datasets (Imagenette)

Is it okay if I start the development of the model script on a separate branch? The current open PR can be merged as soon as it's ready. I believe that there are no more changes to the core functionality of the `imagenet.py` and `tfrecords_utils.py` scripts and they can be assumed frozen for future development. 

Please share your thoughts on the same. 
Thanks.

sayakpaul:
Sounds great and thanks for being so comprehensive about your goals. Should we maintain kanban cards by using the Projects feature of GitHub? @MorganR WDYT? 

For the first three ones, I am game except for a small ask. I'd really like to see a notebook that loads the TFRecords, constructs the data input pipeline, and visualizes a few image batches. 4 and 5 are trivial and I am sure Morgan and I can get you up to speed with those. So, allocating 1/2 days are sufficient IMO. 

> The current open PR can be merged as soon as it's ready. I believe that there are no more changes to the core functionality of the imagenet.py and tfrecords_utils.py scripts and they can be assumed frozen for future development.

I agree. We will only be able to confirm this once we are able to set things up on GCP. For the time being, let's test things with ImageNette because it provides a reasonable representation of what we'd be dealing with. Does that make sense? 

> Is it okay if I start the development of the model script on a separate branch?

I am okay with this. @MorganR WDYT?

AdityaKane2001:
>  I'd really like to see a notebook that loads the TFRecords, constructs the data input pipeline, and visualizes a few image batches. 

Yes, totally. The notebook's almost ready, just making some minor changes.

[...]

PR #3, opened 2021-06-23
Changes in `tfrecords_utils.py` and `imagenet.py`

AdityaKane2001:
@sayakpaul  @MorganR 
Please see [#2](https://github.com/AdityaKane2001/regnety/issues/2#issuecomment-866196972).

One hot encoded, batched the data. `example['synset']` now returns actual label name as opposed to something like `n123456`. This WordNet ID is still available through `example['filename']` as the filename looks like `n123456_123.JPEG`.

Also added `setup.py`. Please see this [comment](https://github.com/AdityaKane2001/regnety/pull/1#issuecomment-863437462). Instead of manually listing down the sub-packages as in the comment, I have used `find_packages()`. This now allows the use of something like 
```python
#/regnety/dataset/imagenet.py
from regnety.regnety.config import config

...
```

Created a new branch because I had started working on the model scripts in the dev branch. 
The only concern is with the numerous `num_parallel_calls`
I will raise an issue in the tensorboard repo as soon as possible.


sayakpaul (approved):

AdityaKane2001:
I will merge this PR after we optimize the input pipeline. I think it will be better that way so that we don't have to go back to this code again and again.

AdityaKane2001:
@sayakpaul @MorganR 

Requesting review. I've made the final changes and everything looks alright. Please share your thoughts on the same.

regnety/dataset/imagenet.py
```diff
@@ -37,8 +36,8 @@ class ImageNet:
     def __init__(
         self,
         tfrecs_filepath: List[str]  = None,
-        batch_size: int = 128,
-        image_size: int = 224,
+        batch_size: int = 1024,
+        image_size: int = 512,
         augment_fn: Union[str, Callable]  = "default",
         num_classes: int = 10,
         randaugment: bool = True,
```
sayakpaul, line 43: `num_classes=10`? Also, we are NOT using the original RandAugment now, should this argument be named as it is? Or is it there to allow the user to incorporate it?

regnety/dataset/imagenet.py
```diff
@@ -48,24 +47,25 @@ def __init__(
...
         self.tfrecs_filepath = tfrecs_filepath
         self.batch_size = batch_size
-        self.image_size = image_size
+        self.image_size = tf.constant([image_size] * self.batch_size) 
         self.augment_fn = augment_fn
         self.num_classes = num_classes
         self.randaugment = randaugment
         if self.randaugment:
-            self._augmenter = RandAugment(magnitude=5, num_layers=2)
+            self.strength = 5
```
sayakpaul, line 55: What does `strength` indicate in our setting?

regnety/dataset/imagenet.py
```diff
@@ -48,24 +47,25 @@ def __init__(
...
             example: A TFRecord dataset example.
-
         Returns:
             Dict containing attributes from a single example. Follows
             the same names as TFRECORDS_FORMAT.
         """
-        image = tf.cast(tf.io.decode_jpeg(example["image"]), tf.float32)
+
+        example =  tf.io.parse_example(example_, _TFRECS_FORMAT)
+        image = tf.reshape(tf.io.decode_jpeg(example["image"]), (512,512,3))
```
sayakpaul, line 68: `512` should be encoded as a variable.

regnety/dataset/imagenet.py
```diff
@@ -80,159 +80,145 @@ def decode_example(self, example: tf.Tensor) -> dict:
...
-            lambda example: tf.io.parse_example(example, _TFRECS_FORMAT),
-            num_parallel_calls = tf.data.AUTOTUNE
-        )
+
+        files = tf.data.Dataset.list_files(self.tfrecs_filepath)
+
+        options = tf.data.Options()
+        options.experimental_deterministic = False
+
+        files = files.with_options(options)
```
sayakpaul, line 95: We can group these. 

Also, when using this and `interleave()` together, does that collectively boost performance? Have you tried using them separately and observed any significant changes?

regnety/dataset/imagenet.py
```diff
@@ -80,159 +80,145 @@ def decode_example(self, example: tf.Tensor) -> dict:
...
+
+        ds = files.interleave(tf.data.TFRecordDataset, 
+          num_parallel_calls = tf.data.AUTOTUNE,
+          deterministic=False)
+
         ds = ds.map(
-            lambda example: self.decode_example(example), 
+            self.decode_example, 
             num_parallel_calls = tf.data.AUTOTUNE
```
sayakpaul, line 103: Let's declare a variable `AUTO = tf.data.AUTOTUNE` and use that.

regnety/dataset/imagenet.py
```diff
@@ -80,159 +80,145 @@ def decode_example(self, example: tf.Tensor) -> dict:
...
+        with keys 'image' and 'target'. 'target' is one hot encoded.
         Args:
-            image: tensor of the image
-            scale_size: Size of image to scale to
-            final_size: Size of final image
-
+            example: an example having keys 'image' and 'label'
         Returns:
-            Tensor of shape (final_size, final_size, 3)
+            example having keys 'image' and 'target'
```
sayakpaul, line 120: If `example` has a special class like `tf.train.Example()` then mention that otherwise `example` -> `Example`.

regnety/dataset/imagenet.py
```diff
@@ -80,159 +80,145 @@ def decode_example(self, example: tf.Tensor) -> dict:
...
-        w = tf.cast(example['width'], tf.int64)
 
 
-        bbox = tf.constant([0.0, 0.0, 1.0, 1.0], 
-                         dtype=tf.float32,
-                         shape=[1, 1, 4])
-        
+    def random_rotate(self, image: tf.Tensor, target: tf.Tensor) -> tuple:
+        """"
+        Randomly rotates images
```
sayakpaul, line 160: Full-stop.

regnety/dataset/imagenet.py
```diff
@@ -80,159 +80,145 @@ def decode_example(self, example: tf.Tensor) -> dict:
...
 
-        Args:
-            example: A dataset example dict.
-            min_area: Minimum area of image to be used
+        Args: 
+            example:example with batch of images to be augmented
 
         Returns:
-            Example of same format as _TFRECS_FORMAT
+            Augmented example with batch of images with same dimensions
```
sayakpaul, line 135: Full-stop.

regnety/profiler.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+from dataset import imagenet
+
+
+
+tf.keras.backend.clear_session()
+
+imgnet = imagenet.ImageNet(
+    tf.io.gfile.glob('gs://adityakane-imagenette-tfrecs/*.tfrecord'),
```
sayakpaul, line 19: This path should come from a config script.

sayakpaul (changes requested):

sayakpaul:
@AdityaKane2001 great work!

A crucial feedback is to go through the doc-strings and make them consistent in terms of style and language. That's about it really.

regnety/profiler.py
AdityaKane2001: Actually profiler.py is a dummy script, just like main.py. I'll delete it when it seems unnecessary. Is that okay?

regnety/dataset/imagenet.py
AdityaKane2001: If you are referring to the usage of `tf.data.Options()`, then no, I didn't see any substantial improvements. But using interleave over plain TFRecordDataset does give some improvements. 

I'll delete the `options` code, as we are specifying deterministic = False in interleave()

regnety/dataset/imagenet.py
AdityaKane2001: Similar to PAWS-TF, here strength influences the delta for color jitter ops.

regnety/dataset/imagenet.py
AdityaKane2001: I'll remove the randaugment argument. No, the user doesn't get the option to implement original RandAugment. 

If augment_fn == 'default' , then we'll apply our augmentations.

regnety/dataset/imagenet.py
sayakpaul: Let's have a comment on that. Also, this is a motivation from SimCLR actually. The strength of augmentation transforms was fairly studied in that paper.

regnety/dataset/imagenet.py
sayakpaul: Cool. That sounds reasonable.

regnety/profiler.py
sayakpaul: Fair enough.

AdityaKane2001:
@sayakpaul I've made the changes.

AdityaKane2001:
@sayakpaul 

About Apache Beam, is [this](https://github.com/swapnil3597/dataflow-tfrecord/blob/master/create_tfrecords/create_tfrecords.py) what we want to do?  

Also, if Beam is quite efficient, do you want me to refactor the existing code as a Beam pipeline? Or will it be an overkill for our purpose?

On a side note, added utility for validation data in imagenet.py and tfrecords_utils.py. Closing #4 now.

sayakpaul:
@AdityaKane2001 [this snippet](https://github.com/swapnil3597/dataflow-tfrecord/blob/master/create_tfrecords/create_tfrecords.py#L185-L192) is actually the meat. I don't think it would require a whole lot of refactoring to be able to plug Apache Beam in. If the refactoring is too much we should dumb it IMO. 

While the snippet you shared uses Dataflow as the pipeline runner, we can use DirectRunner (basically run locally). This is what [the TF-Hub example](https://www.tensorflow.org/hub/tutorials/tf2_semantic_approximate_nearest_neighbors) also follows.

AdityaKane2001:
>  I don't think it would require a whole lot of refactoring 

Sorry, I meant the data loading code, that in [imagenet.py](https://github.com/AdityaKane2001/regnety/blob/temp_dev_data/regnety/dataset/imagenet.py). 

I'm going to implement Beam pipeline for making TFRecords, as discussed in the meet. Just wanted to ask about this.

sayakpaul:
Yeah just doing that is absolutely fine. Otherwise, it might be difficult for the rest of the community to properly follow through with the example. I think this is also an aspect we should consider when we are implementing models at this scale. WDYT?

AdityaKane2001:
I agree. I, too, was baffled at the first glance. 
But I'd really like to keep this as a "priority stretch goal", if that's a thing. Who doesn't want better performance without extra effort?

So for now, I'll add a function and the required classes in `tfrecords_utils.py`, and use them in `make_tfrecords.py`. But later, when time permits, I'll do the same for `imagenet.py`. 

Also, I wanted to talk a bit about the credits scenario. Could you please tell if there is any update about that? Most probably, we'll finish pre-training tasks by Sunday evening, at least the coding part. I'll try to make the TFRecords by the same time. Thus, it would be great if we could start training from next week. The first model will take the most time from my side, but the later ones will be easier. I'll do my homework before that, but the initial delay is something we mostly cannot avoid. 

Please share your thoughts on all of this.

/cc @MorganR

sayakpaul:
> Also, I wanted to talk a bit about the credits scenario. Could you please tell if there is any update about that? Most probably, we'll finish pre-training tasks by Sunday evening, at least the coding part. I'll try to make the TFRecords by the same time. Thus, it would be great if we could start training from next week. The first model will take the most time from my side, but the later ones will be easier. I'll do my homework before that, but the initial delay is something we mostly cannot avoid.

@AdityaKane2001 I think it's sorted now? Cc: @jaeyounkim.

AdityaKane2001:
@sayakpaul 
Yes I got the coupon from Jaeyoun. The credits were redeemed successfully. Thanks for the speedy response!

AdityaKane2001:
@sayakpaul 

Could you please take a look at [tfrecords_utils.py](https://github.com/AdityaKane2001/regnety/blob/db51f75980a5aec66396a8c2ee2e6e31bed95c4c/regnety/utils/tfrecords_utils.py#L341-L420) and [beam_utils.py](https://github.com/AdityaKane2001/regnety/blob/db51f75980a5aec66396a8c2ee2e6e31bed95c4c/regnety/utils/beam_utils.py)?

I executed the pipeline as in `tfrecords_utils.py/make_tfrecs_beam(...)`, but got the following error:

```
WARNING:apache_beam.io.filebasedsink:No shards found to finalize. num_shards: 10, skipped: 10
```

As you will see in the code, I have used `FlatMap`. If I use `ParDo`, I get an error `Example is not iterable`. Could you please take a look and tell if anything's wrong?

sayakpaul:
> As you will see in the code, I have used FlatMap. If I use ParDo, I get an error Example is not iterable. Could you please take a look and tell if anything's wrong?

Yes, it needs to _mapped_. From [here](https://github.com/AdityaKane2001/regnety/blob/db51f75980a5aec66396a8c2ee2e6e31bed95c4c/regnety/utils/tfrecords_utils.py#L341-L420), one thing that seems to be missing is batched elements. After `beam.Create(final_list)`, I think we need to batch the elements and then call `FlatMap()`.

sayakpaul:
@AdityaKane2001 consider including emojis when this stuff happens 😂:

image

Be sure to take notes on what worked and what didn't.

AdityaKane2001:
@sayakpaul 
It works now. I batched the dataset and used ParDo on the batches since we can have each TFRecord made on a different CPU thread.

I'll make TFrecs for the whole dataset by today evening. I just need to make sure that the TFRecords thus made are ok.

I think occasionally adding funny emojis to commit messages is something we can all use, especially since things are quite dull and negative due to the pandemic. Yes, I am taking notes wherever necessary. I intend to add this stuff in the blog.

sayakpaul:
> It works now. I batched the dataset and used ParDo on the batches since we can have each TFRecord made on a different CPU thread.
I'll make TFrecs for the whole dataset by today evening. I just need to make sure that the TFRecords thus made are ok.

That sounds like a reasonable plan to me. Any reason why FlatMap or Map didn't work here? In my experience, ParDo is a bit odd though. Also, a couple of early feedback if that helps:

* Maintain single quotes all the time. 
* Add concise comments when necessary. 

Would be also informative to compare the creation performance with and without Beam. We don't need to do it on the full dataset. Doing it using ImageNette is sufficient for benchmarking. 

> I think occasionally adding funny emojis to commit messages is something we can all use, especially since things are quite dull and negative due to the pandemic.

Oh absolutely! 

> Yes, I am taking notes wherever necessary. I intend to add this stuff in the blog.

We definitely should. Thank you.

AdityaKane2001:
@sayakpaul 

I tried this on AI platform notebook and it worked. The time taken is too high, more than twice the current time (2 minutes vs ~5 minutes for ImageNette). 

In that case, should I proceed with the existing method? I believe it will take around 5 hours to make all TFRecords. Perhaps even less, since we'll be running it on a beefier machine.

sayakpaul:
Just so we are clear, on the same machine the creation process without Beam takes 2 minutes to compete? Is that what you meant?

If so, then I think we should discard it. Since Beam is not the focus of this project I would not recommend any further debugging in the matter.

AdityaKane2001:
> Just so we are clear, on the same machine the creation process without Beam takes 2 minutes to compete? Is that what you meant?

Yes. Sourcing these numbers from the AI Plm. notebook. 

> If so, then I think we should discard it. Since Beam is not the focus of this project I would not recommend any further debugging in the matter.

Sure. I'll start making TFRecords now.

AdityaKane2001:
@sayakpaul Since we're not going to use beam anymore, should I remove that code entirely?

sayakpaul:
I think it's fine to keep it 'cause it's non-trivial work. But we can make a note of its usefulness, efficiency, etc. in the README.

AdityaKane2001:
@sayakpaul @MorganR 

Could you please take a look so that  we can merge this? Most of the code is same as before, apart from some additional utils for Beam. 

Also, I finished making training TFRecords. 

Thanks.

regnety/dataset/imagenet.py
```diff
@@ -18,54 +20,77 @@
...
+        This augmentation is inspired by SimCLR (https://arxiv.org/abs/2002.05709). 
+        The strength parameter is set to 5, which controlsthe effect of augmentations.
+    - Random rotate
+    - Random crop and resize
+
+    If `augment_fn` argument is not set to the string "default", it should be set to 
+    a callable object. That callable must take exactly two arguments: `image` and `target`
+    and must return two values corresponding to the same. 
+
+    If `augment_fn` argument is 'val', then the images woll be center cropped to 224x224.
```
sayakpaul, line 35: `will`

regnety/dataset/imagenet.py
```diff
@@ -18,54 +20,77 @@
...
         Returns:
             Dict containing attributes from a single example. Follows
-            the same names as TFRECORDS_FORMAT.
+            the same names as _TFRECS_FORMAT.
         """
-        image = tf.cast(tf.io.decode_jpeg(example["image"]), tf.float32)
+
+        example =  tf.io.parse_example(example_, _TFRECS_FORMAT)
+        image = tf.reshape(tf.io.decode_jpeg(
+            example["image"]), (self.image_size, self.image_size, 3))
```
sayakpaul, line 93: Just want to ensure we are consistent on the use of single-quotes and double-quotes. If not we'd want to ensure it.

sayakpaul (changes requested):
@AdityaKane2001 looks good. Just a few minor requests.

regnety/dataset/imagenet.py
AdityaKane2001: Corrected wherever required

sayakpaul (approved):

AdityaKane2001:
Merging the PR. Thanks for the suggestions and reviews. Apologies for going a bit fast with this one.

Issue #4, opened 2021-06-26
Optimizing input pipeline


AdityaKane2001:
@sayakpaul @MorganR 

The issue with the profiler was resolved about a day ago. Since then, I've been trying to optimize the input pipeline. Following are my observations:

1. Our current input pipeline is very slow. It takes anywhere between 700,000 us to 1,000,000 us to load a batch of 128 images. Probable reasons in the following points. Recommended [here](https://www.tensorflow.org/guide/data_performance_analysis#1_is_your_tfdata_pipeline_producing_data_fast_enough) is 50 us.
2.  RandAugment takes the most time. If that is removed, the average time drops to 250,000 us. 
3. Unfortunately, we cannot vectorize the entire pipeline. All of our preprocessing functions (`randaugment`,`random_sized_crop`,`scale_and_center_crop`) take single images. The tf functions used themselves take only one image at a time. 
4. I was doing all of this on Colab. Perhaps this may be faster on GCP.

Things I tried:

1. Using `interleave` for TFRecords I/O instead of `map`.
2. Using uint8 for images instead of float32.
3. Removing cast ops. There are no cast ops in the preprocessing code now.

I tried removing `num_parallel_calls` in I/O functions but it increased the time taken.

These changes reduced the time by 200,000 us. These changes were based on guides by TensorFlow ([here](https://www.tensorflow.org/guide/data_performance), [here](https://www.tensorflow.org/guide/data_performance_analysis)). I used the methods which were relevant. 

Lastly, is are these times normal for image datasets? I figured since the training will be on TPUs, the input pipeline should be aggressively optimized. Please correct me if I'm wrong. Is there something that I'm completely missing? Frankly, I have a feeling that I'm blindsided by something very obvious.

Attaching herewith the zip file for all logs of my last run before posting this issue, in case you want to take a look at the actual numbers.
[last_run_logs.zip](https://github.com/AdityaKane2001/regnety/files/6719685/last_run_logs.zip)

sayakpaul:
> The issue with the profiler was resolved about a day ago.

What was the problem? 

Following are my suggestions:

* We are parsing images of 512x512 resolution which definitely is introducing added bottleneck. 
* You don't need to operate on float32 tensors. Because we will be including our normalization/scaling layer inside our model itself. Note that ops like `tf.image.resize()` cast your input tensor to float32, so it's good to cast them back to int. 
* Try order-independent loading:
    ```python
    options = tf.data.Options()
[...]

PR #5, opened 2021-06-29
RegNetY model scripts

AdityaKane2001:
Added:

1. blocks.py: Contains basic building blocks (Stem, Stage, Head, etc.) for RegNetY
2. model.py: Initializes model. Will later include script to initialize using ImageNet weights from TFHub.
3. blocks_test.py and model_test.py: Unittests for respective modules. I'm not sure what to include in model_test, and whether it is logical to include it. Please share your thoughts about the same.
4. config.py: configuration for the four RegNets. 

I know I should have kept the PR smaller, but it would have been difficult to further break this down as all modules are interdependent.
For reference, the file structure is now as follows:

```
.
├── regnety/
│   ├── config/
│   │   ├── __init__.py
│   │   ├── config.py
│   │   └── imagenet_synset_to_human.json
│   ├── model/
│   │   ├── __init__.py
│   │   ├── blocks.py
│   │   ├── blocks_test.py
│   │   ├── model.py
│   │   └── model_test.py
│   ├── dataset/...
│   ├── utils/...
│   ├── main.py
│   └── make_tfrecords.py
├── setup.py
├── README.md
├── LICENSE
└── .gitignore
```


sayakpaul:
Should we first resolve the data pipeline-related stuff and then proceed here? @MorganR WDYT? IMO, that'd be better.

MorganR:
Yes, let's do that.

regnety/config/config.py
```diff
@@ -0,0 +1,116 @@
+from dataclasses import dataclass
+from typing import List
+
+ALLOWED_FLOPS = ('200mf', '400mf', '600mf', '800mf')
```
sayakpaul, line 4: What are these?

regnety/config/__init__.py
```diff
@@ -0,0 +1,6 @@
+import os
+os.environ['TF_CPP_MIN_LOG_LEVEL'] = '3' 
+
+
+from regnety.regnety.config.config import get_model_config, ALLOWED_FLOPS
```
sayakpaul, line 5: Do you have a script for counting flops? If you used an external utility do include an example in a script somewhere in the repository to make things transparent.

regnety/config/config.py
```diff
@@ -0,0 +1,116 @@
...
+    wa: int
+    w0: int
+    wm: float
+
+
+
+def get_model_config(flops: str):
+    """
+    Getter function for configuration for a specific RegNetY model instance.
+    User must provide flops in string format. Example "200MF", "800MF"
```
sayakpaul, line 41: Full-stop.

regnety/config/config.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+
+def get_model_config(flops: str):
+    """
+    Getter function for configuration for a specific RegNetY model instance.
+    User must provide flops in string format. Example "200MF", "800MF" 
+    
+    Args: 
+        flops: String denoting flops of model (as per the paper)
```
sayakpaul, line 44: Provide the arXiv link.

regnety/config/config.py
```diff
@@ -0,0 +1,116 @@
...
+            wm = 2.5
+        )
+        
+    elif flops.lower() == '400mf':
+        return RegNetYConfig(
+            name = "RegNetY 400MF",
+            flops = "400MF",
+            num_classes = 1000,
+            depths = [1, 3, 3, 6],
+            widths = [48, 104, 208, 440],
```
sayakpaul, line 79: This relationship was derived from the explorations conducted in the paper. A linear quantized relationship if I am remembering correctly. Do we want to note this at the top? Just a single line would suffice.

regnety/models/blocks.py
```diff
@@ -0,0 +1,219 @@
+"""Has building blocks of RegNetY models"""
```
sayakpaul, line 1: How about `"""Contains the building blocks of RegNetY models."""`?

regnety/models/blocks.py
```diff
@@ -0,0 +1,219 @@
...
+# 2. Body
+#     2.1 Block
+#         2.1.1 SE
+# 3. Head
+# 4. RegNetY
+# Reference: https://tinyurl.com/dcc5c5p8
+
+class Stem(tf.keras.layers.Layer):
+    """Class to initiate stem architecture from the paper: 
+    `stride-two 3×3 conv with w0 = 32 output filters`
```
sayakpaul, line 17: Full-stop and arXiv link.

regnety/models/blocks.py
```diff
@@ -0,0 +1,219 @@
...
+    
+    Args:
+        None, stem is common to all models
+    """
+
+    def __init__(self):
+        super(Stem, self).__init__()
+        self.conv3x3 =  tf.keras.layers.Conv2D(32, (3,3), strides = 2)
+        self.bn = tf.keras.layers.BatchNormalization(
```
sayakpaul, line 26: Let's import `layers`. Also, let's provide names for each of the layers which is a standard practice for models at this scale inside of Keras.

regnety/models/blocks.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def call(self, inputs):
+        x = self.conv3x3(inputs)
+        x = self.bn(x)
+        x = self.act(x)
+        return x
+
+class SE(tf.keras.layers.Layer):
+    """
+    Squeeze and Excite block. Takes se_ratio and in_filters as arguments.
```
sayakpaul, line 39: Provide the arXiv link of the SE paper. Isn't it "Squeeze and Excitation"? 

SE blocks tend to contribute to the increased amount of latency (primarily because of the way the Dense layers are placed inside of it). [GC Blocks](https://arxiv.org/pdf/1904.11492) provide a nice way around. Maybe we can work this out in future.

regnety/models/blocks.py
```diff
@@ -0,0 +1,219 @@
...
+        self.ga_pool = tf.keras.layers.GlobalAveragePooling2D()
+        self.squeeze_dense = tf.keras.layers.Dense(self.se_filters, activation = 'relu')
+        self.excite_dense = tf.keras.layers.Dense(self.out_filters,  activation = 'sigmoid')
+        
+    def call(self, inputs):
+        # input shape: (h,w,out_filters)
+        x = self.ga_pool(inputs) #x: (out_filters)
+        x = self.squeeze_dense(x) #x: (se_filters)
+        x = self.excite_dense(x) #x: (out_filters)
+        x = tf.reshape(x, [-1,1,1,self.out_filters])
```
sayakpaul, line 63: Space between `#` and what follows it.

regnety/models/blocks.py
```diff
@@ -0,0 +1,219 @@
...
+        x = tf.math.multiply(x, inputs) #x: (h,w,out_filters)
+        return x
+        
+
+class YBlock(tf.keras.layers.Layer):
+    """
+    Y Block in RegNetY structure. 
+    IMPORTANT: Grouped convolutions are only supported by keras on GPU.
```
sayakpaul, line 71: > Grouped convolutions are only supported by keras on GPU. 

Is this still true?

regnety/models/blocks.py
```diff
@@ -0,0 +1,219 @@
...
+        
+        x = x + skip_tensor
+
+        return x
+    
+
+
+class Stage(tf.keras.layers.Layer):
+    """
+    Class for RegNetY stage.
```
sayakpaul, line 163: Add another sentence to denote what stage does. UX-wise it's helpful.

regnety/models/blocks.py
```diff
@@ -0,0 +1,219 @@
...
+    Head for all RegNetY models.
+
+    Args:
+        num_classes: Integer specifying number of classes of data. 
+    """
+    def __init__(self, num_classes):
+        super(Head, self).__init__()
+
+        self.gap = tf.keras.layers.GlobalAveragePooling2D()
+        self.fc = tf.keras.layers.Dense(num_classes, activation = 'sigmoid')
```
sayakpaul, line 211: `sigmoid` for a multi-class classification problem? Shouldn't this be softmax? Also, let's separate out the last layer activation a distinct layer to give the users a chance to operate on the raw logits.

regnety/models/blocks_test.py
```diff
@@ -0,0 +1,63 @@
+import os
+os.environ['TF_CPP_MIN_LOG_LEVEL'] = '3'
```
sayakpaul, line 2: Add a space after this.

regnety/models/blocks_test.py
```diff
@@ -0,0 +1,63 @@
...
+            tf.keras.layers.GlobalAveragePooling2D(),
+            tf.keras.layers.Dense(10, activation= 'sigmoid')
+            ])
+        model.compile(optimizer = 'adam',
+            metrics=['accuracy'], loss = 'categorical_crossentropy')
+        model.fit(self.image,self.y,verbose=0)
+
+    def test_YBlock_stride1(self):
+        if len(self.devices) > 1:
+            #Grouped convs are available only on GPU
```
sayakpaul, line 30: Space.

regnety/models/blocks_test.py
```diff
@@ -0,0 +1,63 @@
...
+            model.fit(self.image,self.y,verbose=0)
+        else:
+            raise unittest.case.SkipTest("GPU not available on this machine, skipping.")
+
+
+
+
+if __name__ == '__main__':
+    unittest.main()
+
```
sayakpaul, line 63: For unit tests, how about the following?

* Take dummy inputs and dummy labels. 
* Verify if the output shapes of the individual blocks are as expected. 

We can probably leave out the unit-testing for the model as it's not that well-defined and understood IMO.

sayakpaul (changes requested):
@AdityaKane2001 excellent work here. I just wanted to have my reviews to you beforehand to avoid any blockers. But please don't feel like working on them yet.

AdityaKane2001:
> @AdityaKane2001 excellent work here. I just wanted to have my reviews to you beforehand to avoid any blockers. But please don't feel like working on them yet.

Thanks a lot. Actually I had requested the reviews for that exact reason, so you could review when it's convenient for you, and I'll work on them along with the input pipeline.

regnety/models/blocks.py
AdityaKane2001: I'm not aware of any new development regarding that. AFAIK, they're not supported on CPU. I'm sure that they work on GPU since I've tested that, but I'm not sure about TPU. I'll check that and update accordingly.

sayakpaul: I had provided a snippet. Did you try that?

regnety/config/__init__.py
AdityaKane2001: No I don't have one currently. I'll work that out shortly.

regnety/models/blocks.py
AdityaKane2001: I think we agreed on using the Keras implementation itself. ([comment](https://github.com/AdityaKane2001/regnety/issues/4#issuecomment-870201961))
No I haven't filed an issue yet. I'll do that. Here's some additional info I found on this. Please see [this](https://github.com/tensorflow/tensorflow/blob/35c46d26cf420e7076abdcc3324f7a8b80f4710a/tensorflow/core/kernels/conv_ops.cc#L144-L226). Grouped convs are implemented in TF (albeit with bad performance) on CPU but they may not necessarily be a part of Keras, just as they were part of core TF before being added to Keras ([reference](https://github.com/tensorflow/tensorflow/pull/36773)).

regnety/models/blocks.py
AdityaKane2001: Sure, that's what they did in the official implementation.

sayakpaul: Yeah but the `sigmoid` part is definitely not correct unless I am missing out on something.

regnety/models/blocks.py
sayakpaul: Oh cool.

regnety/models/blocks.py
AdityaKane2001: > Or we can omit the softmax layer totally and specify from_logits=True in Categorical_Crossentropy().

I was replying to this. Sigmoid is incorrect, I agree.

regnety/models/blocks.py
AdityaKane2001: @sayakpaul 

Should I go for GC blocks right away, or should we keep it as a stretch goal?

sayakpaul: Stretch.

AdityaKane2001:
@sayakpaul  

These scripts are good to go from my side. @MorganR  could you please take a look? 
Thanks

regnety/config/config.py
```diff
@@ -0,0 +1,121 @@
...
+ALLOWED_FLOPS = ('200mf', '400mf', '600mf', '800mf')
+
+@dataclass
+class RegNetYConfig:
+    """
+    Dataclass for architecture configuration for RegNetY.
+
+    Args:
+        name: Name of the model eg. "RegNetY200MF"
+        flops: Flops of the model eg. "400MF"
```
sayakpaul, line 13: Put a short sentence in brackets to denote what 400MF means for clarity.

regnety/models/model.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def call(self, inputs):
+        self.model.call(inputs)
+    
+    def get_model(self):
+        """Returns sequential model constructed in this class"""
+        return self.model
+    
+
+    def _get_model_with_config(self, config):
```
sayakpaul, line 37: Pretty cool to see what one can achieve even with `Sequential`. Great work framing the blocks in such a manner.

regnety/models/model.py
```diff
@@ -0,0 +1,69 @@
...
+        """
+        Makes a tf.keras.Sequential model using the given config.
+         
+        """
+        model = tf.keras.models.Sequential()
+        model.add(tf.keras.layers.InputLayer(input_shape = (224,224,3)))
```
sayakpaul, line 43: Should we get this shape from a configuration file? I'd also suggest looking for the `keras.applications` classes since they allow variable-shape images during inference. If we prefix this here then that possibility gets nullified. I think allowing a model to do inference on shapes not discovered during training is quite practical.

WDYT?

regnety/models/model.py
```diff
@@ -0,0 +1,69 @@
...
+        model.add(PreStem())
+        model.add(Stem())
+
+        in_channels = 32 # Output channels from Stem
+
+        for i in range(4):  # 4 stages
+            depth = config.depths[i]
+            out_channels = config.widths[i]
+            group_width = config.group_width
```
sayakpaul, line 52: Can't we determine these based on the FLOP config received?

sayakpaul (changes requested):

regnety/config/config.py
```diff
@@ -0,0 +1,121 @@
...
+            num_classes = 1000,
+            depths = [1, 1, 4, 7],
+            widths = [24, 56, 152, 368],
+            group_width = 8,
+            bottleneck_ratio = 1,
+            SE_ratio = 0.25,
+            wa = 36,
+            w0 = 24,
+            wm = 2.5
+        )
```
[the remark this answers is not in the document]
AdityaKane2001: I think this approach is marginally better, since we don't create multiple unnecessary objects of `RegnetYConfig`. But I can do dict if you want.

regnety/models/blocks.py
```diff
@@ -0,0 +1,353 @@
...
+    """
+
+    def __init__(self,
+        mean: tf.Tensor = _MEAN,
+        variance: tf.Tensor = _VAR,
+    ):
+        super(PreStem, self).__init__(name = 'PreStem')
+        self.mean = mean
+        self.var = variance
+        # self.inplayer = tf.keras.layers.InputLayer(input_shape = (224,224,3))
```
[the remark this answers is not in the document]
AdityaKane2001: Sorry, forgot to uncomment after debugging.

regnety/models/blocks.py
```diff
@@ -0,0 +1,353 @@
...
+
+        self.resize = layers.experimental.preprocessing.Resizing(
+            224, 224, name = 'prestem_resize'
+        )
+        self.norm = layers.experimental.preprocessing.Normalization(
+            mean = self.mean, variance = self.var, name = 'prestem_normalize'
+        )
```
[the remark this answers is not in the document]
AdityaKane2001: In the input pipeline. Please see regnety/rengnety/dataset/imagenet.py (in temp_dev_data branch).

regnety/config/config.py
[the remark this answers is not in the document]
sayakpaul: I will go with what you had envisioned.

regnety/models/model.py
AdityaKane2001: Could you elaborate on that? `in_channels` and range(4) are defined from the paper itself.

regnety/models/model.py
AdityaKane2001: Sure. I'll add an `input_shape` arg in \_\_init\_\_ and make sure it is >= 224. If not specified, will default to 224. I believe this is the gist of what they've done in keras.applications.

sayakpaul: Yes, sounds good to me.

regnety/models/model.py
```diff
@@ -0,0 +1,90 @@
...
+    def __init__(self, flops:str = "", input_shape: Union[List, Tuple] = None ):
+        super(RegNetY, self).__init__()
+
+        if flops not in ALLOWED_FLOPS:
+            raise ValueError("`flops` must be one of " + str(ALLOWED_FLOPS))
+        
+        if input_shape is None:
+            self.input_shape = (224,224,3)
+        else: 
+            self.input_shape = input_shape
```
sayakpaul, line 30: It's causing a conflict with a read-only @property of the base class.

regnety/models/model.py
```diff
@@ -0,0 +1,90 @@
...
+            raise ValueError('Input shape is invalid. Please enter input shape '
+                             ' as (height, width, 3)')
+
+
+        self.flops = flops
+        self.config = get_model_config(self.flops)
+        self.model = self._get_model_with_config(self.config)
+
+    def build(self, input_shape):
+        self.model.build(input_shape)
```
sayakpaul, line 48: I think we should use the input shape provided or set during initialization here.

regnety/models/model.py
```diff
@@ -0,0 +1,90 @@
...
+            ))
+            
+            in_channels = out_channels
+        
+        model.add(Head(config.num_classes))
+
+        return model
+
+model = RegNetY('200mf')
+print([layer.name for layer in model.get_layer('sequential').layers])
```
sayakpaul, line 90: Be sure to remove these two lines of code.

sayakpaul (changes requested):

regnety/models/model.py
AdityaKane2001: Sorry, forgot about that. I'll change it.

regnety/models/model.py
AdityaKane2001: I don't think we should do that. If the input shape provided by the user is incorrect, then we must rely on the shape of the inputs. We can use this shape for InputLayer. Also, I'm not sure about the behavior if we use the user defined shape.

sayakpaul: If this is the case, then are we letting the users initialize the model with the argument `input_shape`?

AdityaKane2001: Is it okay if we discuss this in tomorrow's meet? I have a couple of questions related to this.

regnety/models/model.py
```diff
@@ -0,0 +1,74 @@
...
+        ))
+
+        in_channels = out_channels
+
+    model.add(Head(config.num_classes))
+
+    return model
+
+
+def RegNetY(flops: str = "", input_shape: Union[List, Tuple] = None):
```
MorganR, line 42: It would be nice to indicate that this returns a tf.keras.Model (with type annotations).

regnety/models/model.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    Instantiates a RegNetY instance based on flops and input_shape furnished by user.
+   
+    Args:
+        flops: Flops of the model eg. "400MF" (Processing one image requires 
+            400 million floating point operations (multiplication, addition))
+        input_shape: A python list or tuple denoting the shape input. Please omit the 
+            batch dimension. eg. (256, 256, 3). Must be greater than or equal to 224.
```
MorganR, line 50: Please include a Returns: section as well

regnety/models/model.py
```diff
@@ -0,0 +1,74 @@
...
+                            ' must be greater than or equal to 224.')
+
+    try:
+        assert len(userdef_input_shape) == 3
+    except:
+        raise ValueError('Input shape is invalid. Please enter input shape '
+                            ' as (height, width, 3)')
+
+
+    flops = flops
```
MorganR, line 71: nit: is this necessary?

MorganR (approved):

sayakpaul:
Configuring them inside `__init__.py` of the `regnety` module? @MorganR might have a more clever idea here.

AdityaKane2001:
> Configuring them inside __init__.py of the regnety module? 

I tried this by adding an init file ./regnety/\_\_init\_\_.py:

```python
from regnety.regnety import models, dataset, config

```
But this didn't have any effect.

sayakpaul:
I'd not worry about it really.

AdityaKane2001:
@sayakpaul 
If there's nothing else, should I merge this PR?

sayakpaul:
Sure. I guess there are some conflicts that are needed to be resolved first.

AdityaKane2001:
Merging this one. Thanks for your reviews and suggestions.

AdityaKane2001:
@sayakpaul 
Could you please approve this? I'm not allowed to merge since you had requested changes. Thanks.

sayakpaul (approved):

PR #11, opened 2021-07-17
Added training utilities

AdityaKane2001:
Added:
- train.py
- train_utils.py
- config.py : get_train_config and get_custom_train_comfig
- utility in imagenet.py for splitting dataset into two. 
- option to exclude `color_jitter` augmentation


regnety/config/config.py
```diff
@@ -52,71 +84,121 @@ def get_model_config(flops: str):
...
+        optimizer="sgd",
+        base_lr=0.8,
+        warmup_epochs=5,
+        warmup_factor=0.1,
+        total_epochs=100,
+        weight_decay=5e-4,
+        momentum=0.9,
+        lr_schedule="half_cos",
+        log_dir="gs://adityakane-train/logs",
+        model_dir="gs://adityakane-train/models"
```
sayakpaul, line 178: Should we name it more appropriately? Something like `gs://in1k-regnety-*`?

regnety/dataset/imagenet.py
```diff
@@ -49,6 +52,8 @@ def __init__(
         image_size: int = 512,
         augment_fn: Union[str, Callable]  = "default",
         num_classes: int = 1000,
+        percent_valid: int = 1,
+        color_jitter: bool = False, 
         scale_to_unit: bool = True
```
sayakpaul, line 57: Space in between arguments and values.

regnety/dataset/imagenet.py
```diff
@@ -60,6 +65,8 @@ def __init__(
         self.image_size = image_size
         self.augment_fn = augment_fn
         self.num_classes = num_classes
+        self.percent_valid = percent_valid
+        self.color_jitter = color_jitter
         self.scale_to_unit = scale_to_unit
         
         if self.augment_fn == "default":
```
sayakpaul, line 72: Space in between arguments and values.

regnety/dataset/imagenet.py
```diff
@@ -127,14 +134,37 @@ def _read_tfrecs(self) -> Type[tf.data.Dataset]:
             num_parallel_calls = AUTO 
         )
 
-        ds = ds.cache("tf_cache")
-       
-        ds = ds.repeat()
-        ds = ds.batch(self.batch_size)
+        ds = ds.cache("gs://adityakane-train/cache/"+str(time.time()))
```
sayakpaul, line 137: Let's get the caching address from config and use `os.path.join()`?

regnety/dataset/imagenet.py
```diff
@@ -127,14 +134,37 @@ def _read_tfrecs(self) -> Type[tf.data.Dataset]:
...
+        
+        ds = ds.batch(self.batch_size, drop_remainder=True)
```
sayakpaul, line 139: Why the remainder is being dropped?

regnety/dataset/imagenet.py
```diff
@@ -162,10 +192,25 @@ def color_jitter(self, image: tf.Tensor, target: tf.Tensor) -> tuple:
...
+
+        Args: 
+            image: Batch of images to perform random rotation on.
+            target: Target tensor.
+
+        Returns:
+            Augmented example with batch of images and targets with same dimensions.
+        """
+
+        aug_images = tf.image.random_flip_left_right(image)
```
sayakpaul, line 212: Can't we map this directly? Why a separate function for it?

regnety/dataset/imagenet.py
```diff
@@ -214,6 +259,7 @@ def center_crop_224(self, image: tf.Tensor, target: tf.Tensor) -> tuple:
         """
         aug_images = tf.image.resize(image, (320, 320))
         aug_images = tf.image.central_crop(aug_images, 224./320.)
+        aug_images = tf.image.resize(image, (320, 320))
```
sayakpaul, line 262: Prefer having these values in configurations. That way it's comprehensive to the users.

regnety/dataset/imagenet.py
```diff
@@ -262,14 +308,27 @@ def make_dataset(self) -> Type[tf.data.Dataset]:
...
+            self._one_hot_encode_example,
+            num_parallel_calls=AUTO
+        )
+
         if self.default_augment:
+            if self.color_jitter:
+                ds = ds.map(
+                    self.color_jitter,
+                    num_parallel_calls = AUTO
+                )
```
sayakpaul, line 328: Spaces in between args and their values.

regnety/make_tfrecords.py
```diff
@@ -7,12 +7,12 @@ def main():
...
-    parser.add_argument("-synpath","--synset_filepath", type=str, default='')
+    parser.add_argument("-synpath","--synset_filepath", type=str, default="")
     parser.add_argument("-bs", "--batch_size", type=int, default=1024)
     parser.add_argument("-f", "--log_freq",type=int, default=50)
-    parser.add_argument("-s", "--shuffle", action='store_true')
-    parser.add_argument("-v", "--validation_set",action='store_true')
-    parser.add_argument("-b", "--use_apache_beam",action='store_true')
+    parser.add_argument("-s", "--shuffle", action="store_true")
+    parser.add_argument("-v", "--validation_set",action="store_true")
+    parser.add_argument("-b", "--use_apache_beam",action="store_true")
```
sayakpaul, line 15: Would suggest adding `help` for the arguments too.

sayakpaul (changes requested):

regnety/config/config.py
AdityaKane2001: If it's okay, I could make another directory in the bucket, for the final models.

regnety/dataset/imagenet.py
AdityaKane2001: Didn't follow for args with type hints due to [this](https://google.github.io/styleguide/pyguide.html#36-whitespace).

regnety/dataset/imagenet.py
AdityaKane2001: Sorry, but these are assignments.

sayakpaul: Ah my bad.

regnety/dataset/imagenet.py
sayakpaul: Ah I see, yes. Makes sense.

regnety/config/config.py
sayakpaul: Okay.

regnety/dataset/imagenet.py
AdityaKane2001: I added that we're not implementing `steps_per_epoch` for now. I'll change this once we fix that.

regnety/dataset/imagenet.py
AdityaKane2001: We can, but that'll require a lambda function. No technical reason, I just wanted avoid clutter in `make_dataset()`.

regnety/dataset/imagenet.py
AdityaKane2001: Actually, there is no other function that requires information a config file. Is it okay of I include `crop_size` argument in the constructor, and set it to a default value? 
I prefer if the input pipeline takes no config file, so that it can be used independently. But please tell me if it's better otherwise, I'll change it.

sayakpaul: I think having a separate configuration file for the input pipeline would make it more customizable.

regnety/dataset/imagenet.py
AdityaKane2001: Actually, this doesn't work as expected, since `os.path.join` uses "/" as a separator. It created a chain of directories inside the bucket. (gs: `/` ` / ` adityakane-train `/`... )

regnety/dataset/imagenet.py
AdityaKane2001: Sure, no problem. I'll make a dataclass for that.

regnety/dataset/imagenet.py
sayakpaul: I meant something like `os.path.join("gs://adityakane-train/cache", ...)`.

AdityaKane2001:
@sayakpaul 

I've implemented the additional config dataclass. Also, I've added the option to scale to (-1,1) or (0,1) as in keras applications. Please take a look.

AdityaKane2001:
@MorganR  @sayakpaul 
Could you please take a look at this one?

regnety/dataset/imagenet.py
```diff
@@ -198,6 +208,7 @@ def random_crop(self, image: tf.Tensor, target: tf.Tensor) -> tuple:
         """
 
         cropped = tf.image.random_crop(image, size = (self.batch_size, 320, 320, 3))
+        cropped = tf.image.resize(cropped, size=(224,224))
         return cropped, target
     
     def center_crop_224(self, image: tf.Tensor, target: tf.Tensor) -> tuple:
```
MorganR, line 214: This name no longer makes sense, since the crop size may or may not be 224, depending on the config.

regnety/config/config.py
```diff
@@ -52,71 +148,124 @@ def get_model_config(flops: str):
...
+
+
+def get_train_config():
+    """
+    Getter function for training config. Config is same as in the paper
+    (see link above). If ambiguous, 
+    https://github.com/facebookresearch/pycls/blob/master/pycls/core/config.py  
+    is assumed to be source of truth.
+
+    Args: None
```
MorganR, line 226: Please update

regnety/config/config.py
```diff
@@ -35,12 +38,105 @@ class RegNetYConfig:
...
+    Returns:
+        A PreprocessingConfig dataclass instance with all attributes
+    
+    """
+    return PreprocessingConfig(
+        tfrecs_filepath=tfrecs_filepath,
+        batch_size=batch_size,
+        image_size=image_size,
+        crop_size=crop_size,
+        resize_to_size=resize_to_size,
```
MorganR, line 124: Please clarify in the Pydoc why both crop_size and resize_to_size are needed, and what the difference is.

regnety/utils/train_utils.py
```diff
@@ -0,0 +1,124 @@
...
+        def half_cos_schedule(epoch, lr):
+            # Taken from pycls/pycls/core/optimizer.py, since not clear from paper.
+            if epoch < cfg.warmup_epochs:
+                new_lr = 0.5 * (1.0 + tf.math.cos(PI * epoch /
+                    cfg.total_epochs)) * cfg.base_lr
+                alpha = epoch / cfg.warmup_epochs
+                warmup_factor = cfg.warmup_factor * (1. - alpha) + alpha
+                return new_lr * warmup_factor
+            else:
+                new_lr = new_lr = 0.5 * (1.0 + tf.math.cos(PI * epoch /
```
MorganR, line 47: nit: remove one `new_lr = `

MorganR (changes requested):

regnety/dataset/imagenet.py
```diff
@@ -198,6 +208,7 @@ def random_crop(self, image: tf.Tensor, target: tf.Tensor) -> tuple:
         """
 
         cropped = tf.image.random_crop(image, size = (self.batch_size, 320, 320, 3))
+        cropped = tf.image.resize(cropped, size=(224,224))
```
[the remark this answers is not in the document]
AdityaKane2001: Actually one resize call is in the model, in PreStem class. So I'll remove this call from here. 

So now only center_crop uses resize, which is to be called before taking the central crop. Thus, we cannot remove it. 

So, if I'm not wrong, for each batch, resize gets called atmost twice: once in center_crop (if it's a val ds) and once in the model.

regnety/config/config.py
```diff
@@ -52,71 +148,124 @@ def get_model_config(flops: str):
...
+
+    Args: None
+
+    Returns:
+        A TrainConfig dataclass instance.
+    """
+
+    return TrainConfig(
+        optimizer="sgd",
+        base_lr=0.8,
```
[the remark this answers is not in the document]
AdityaKane2001: This LR was taken from pycls, but is irrelevant now since we're going to use Adam. In hindsight, I think it would be better to just use `get_custom_train_config` all the time because it's more explicit.
Please share your thoughts on the same.

regnety/config/config.py
AdityaKane2001: I'm sorry, I didn't get what should be updated. Could you please elaborate?

MorganR: Sorry, please ignore.

regnety/config/config.py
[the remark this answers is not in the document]
MorganR: Sounds good.

regnety/config/config.py
MorganR: Thanks for explaining this. I think `resize_pre_crop` would be a better name in that case.

MorganR (changes requested):

regnety/config/config.py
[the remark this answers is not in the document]
AdityaKane2001: Removed `get_custom_train_config`. Now `get_train_config` has default keyword arguments.

AdityaKane2001:
@MorganR  
Made requested changes

AdityaKane2001:
@sayakpaul 

Could you please review this? It would be great if we could merge this soon, since we're starting training.

regnety/config/config.py
```diff
@@ -35,12 +38,109 @@ class RegNetYConfig:
...
+
+def get_preprocessing_config(
+    tfrecs_filepath: List[str] = None,
+    batch_size: int = 1024,
+    image_size: int = 512,
+    crop_size: int = 224,
+    resize_pre_crop: int = 320,
+    augment_fn: Union[str, Callable] = "default",
+    num_classes: int = 1000,
+    cache_dir: str = "gs://ak-europe-train/cache/",
```
sayakpaul, line 96: Since we are disabling caching (and it makes sense because of the volume of the dataset) I think it's better to remove this one to clear any confusion.

regnety/config/config.py
```diff
@@ -52,71 +152,108 @@ def get_model_config(flops: str):
...
+    base_lr: float = 0.001 * 8,
+    warmup_epochs: int = 5,
+    warmup_factor: float = 0.1,
+    total_epochs: int = 100,
+    weight_decay: float = 5e-5,
+    momentum: float = 0.9,
+    lr_schedule: str = "half_cos",
+    log_dir: str = "gs://ak-europe-train/logs",
+    model_dir: str = "gs://ak-europe-train/models",
+    cache_dir: str = "gs: // ak-europe-train/cache"
```
sayakpaul, line 233: Remove this one maybe?

regnety/dataset/imagenet.py
```diff
@@ -227,10 +239,17 @@ def _scale_to_unit(self, image: tf.Tensor, target: tf.Tensor) -> tuple:
...
+            aug_images = tf.cast(image, tf.float32)
+            aug_images = aug_images / 127.5
+            aug_images = aug_images - 1.
+            return aug_images, target
+        
+        else:
+            aug_images = tf.cast(image, tf.float32)
+            aug_images = aug_images / 255.
+            return aug_images, target
+
```
sayakpaul, line 252: Aren't we doing this inside the model? And if so, I think it's better to remove it.

regnety/make_tfrecords.py
```diff
@@ -4,15 +4,24 @@
...
-    parser.add_argument("-synpath","--synset_filepath", type=str, default='')
-    parser.add_argument("-bs", "--batch_size", type=int, default=1024)
-    parser.add_argument("-f", "--log_freq",type=int, default=50)
-    parser.add_argument("-s", "--shuffle", action='store_true')
-    parser.add_argument("-v", "--validation_set",action='store_true')
-    parser.add_argument("-b", "--use_apache_beam",action='store_true')
+    parser.add_argument("-o", "--odir", type=str,
+        help="Output directory for TFRecords.")
+    parser.add_argument("-d", "--data_dir", type=str,
+        help="Input data directory. Must have ImageNet-like directory structure.")
```
sayakpaul, line 10: If we could provide a link to how ImageNet-like folder structure looks like it'd be great. But don't sweat it.

regnety/models/model.py
```diff
@@ -7,7 +7,7 @@
 
 
 def _get_model_with_config(config,  
-    userdef_input_shape: Union[List, Tuple] = (224,224,3)):
+    userdef_input_shape: Union[List, Tuple] = (320,320,3)):
```
sayakpaul, line 10: Are we training with 320x320 now? Or is there such that our input size becomes 320 -> random_crop -> resize to 224?

regnety/train.py
```diff
@@ -0,0 +1,109 @@
...
+val_prep_cfg = get_preprocessing_config(
+    tfrecs_filepath=val_tfrecs_filepath,
+    augment_fn="val"
+)
+
+print("Training options detected:", train_cfg)
+print("Preprocessing options detected.")
+print("Training on TFRecords: ",
+    train_prep_cfg.tfrecs_filepath[0], " to ",train_prep_cfg.tfrecs_filepath[-1])
+print("Validating on TFRecords: ",
```
sayakpaul, line 74: Let's log this using the standard `logger` module.

regnety/tuner.py
```diff
@@ -0,0 +1,110 @@
+from regnety.regnety.config.config import (
```
sayakpaul, line 1: I think it's okay to keep this script for reference here. However, I'd also make a brief note here about its usage i.e. we tuned hyperparameters using ImageNette to get a rough sense of the best set of hyperparameters to start our experiments with.

sayakpaul (changes requested):

regnety/config/config.py
AdityaKane2001: I kept it so it can be used for smaller datasets, specifically ImageNette. But makes sense, I'll remove it.

regnety/dataset/imagenet.py
AdityaKane2001: No, 'cause there's no Rescaling layer in TF Keras 2.5. It is present only in the nightly version, I think. THus, it needs to be done in preprocessing pipeline itself.

sayakpaul: I am confused by this one: https://www.tensorflow.org/api_docs/python/tf/keras/layers/experimental/preprocessing/Rescaling. It's 2.5.0. In fact, it was there since 2.4.1. Or am I missing something?

regnety/models/model.py
AdityaKane2001: That's correct. We have 512x512 images, we random crop them to 320x320 and pass them to the model. The PreStem layer resizes them to 224x224.

regnety/tuner.py
AdityaKane2001: >  brief note here about its usage

Do we want to note this at the beginning of the module or the docs?

regnety/dataset/imagenet.py
AdityaKane2001: Sorry, I was misled by [this](https://user-images.githubusercontent.com/64411306/126889024-f1bc2582-f8c2-489b-ad2d-3e8c732d7f38.png). I'll change that asap.

regnety/tuner.py
sayakpaul: Noting this at the beginning of the script sounds good to me.

regnety/dataset/imagenet.py
AdityaKane2001: In this case, do we want to keep just one of them, the /255. one? Or should I include both?

sayakpaul: If we are mean-subtracting the channels then /255 is anyway required to make the value range suitable. Even if the mean-subtraction part is omitted just scaling pixel values to [0, 1] is still a widely used practice I'd say.

AdityaKane2001:
@sayakpaul

Made requested changes.

regnety/config/config.py
```diff
@@ -35,12 +38,103 @@ class RegNetYConfig:
...
+    tfrecs_filepath: List[str]
+    batch_size: int
+    image_size: int
+    crop_size: int
+    resize_pre_crop: int
+    augment_fn: Union[str, Callable]
+    num_classes: int
+    color_jitter: bool
+    scale_to_unit: bool
+    scale_method: str
```
sayakpaul, line 82: Will this be required now?

regnety/config/config.py
```diff
@@ -52,71 +146,106 @@ def get_model_config(flops: str):
...
-            bottleneck_ratio = 1,
-            SE_ratio = 0.25,
-            wa = 39,
-            w0 = 56,
-            wm = 2.4
+            name="RegNetY 800MF",
+            flops="800MF",
+            num_classes=1000,
+            depths=[1, 3, 8, 2],
+            widths=[64, 128, 320, 768],
```
sayakpaul, line 207: Apologies if I forgot. But have we mentioned that these values were derived in the paper and essentially correspond to a quantized linear function?

regnety/config/config.py
```diff
@@ -52,71 +146,106 @@ def get_model_config(flops: str):
...
+            group_width=16,
+            bottleneck_ratio=1,
+            SE_ratio=0.25,
+            wa=39,
+            w0=56,
+            wm=2.4
```
sayakpaul, line 213: Could we go for more descriptive variable names? Also, why is `SE_ratio` in such casing?

regnety/dataset/imagenet.py
```diff
@@ -35,32 +36,25 @@ class ImageNet:
...
-        if (tfrecs_filepath is None) or  (tfrecs_filepath == []):
+        self.tfrecs_filepath = cfg.tfrecs_filepath
+        self.batch_size = cfg.batch_size
+        self.image_size = cfg.image_size
+        self.crop_size = cfg.crop_size
+        self.resize_pre_crop = cfg.resize_pre_crop
+        self.augment_fn = cfg.augment_fn
+        self.num_classes = cfg.num_classes
+        self.color_jitter = cfg.color_jitter
+        self.scale_method = cfg.scale_method
```
sayakpaul, line 54: Spaces between arguments and their values.

regnety/make_tfrecords.py
```diff
@@ -4,15 +4,24 @@
...
-    parser.add_argument("-synpath","--synset_filepath", type=str, default='')
-    parser.add_argument("-bs", "--batch_size", type=int, default=1024)
-    parser.add_argument("-f", "--log_freq",type=int, default=50)
-    parser.add_argument("-s", "--shuffle", action='store_true')
-    parser.add_argument("-v", "--validation_set",action='store_true')
-    parser.add_argument("-b", "--use_apache_beam",action='store_true')
+    parser.add_argument("-o", "--odir", type=str,
+        help="Output directory for TFRecords.")
+    parser.add_argument("-d", "--data_dir", type=str,
+        help="Input data directory. Must have ImageNet-like directory structure as shown in https://gist.github.com/AdityaKane2001/490b8c94c05538dec690513022195e91")
```
sayakpaul, line 10: We can shorten the Gist URL with https://git.io. This would improve visibility.

regnety/models/blocks.py
```diff
@@ -35,6 +35,9 @@ def __init__(self,
         self.mean = mean
         self.var = variance
 
+        self.rescale = layers.experimental.preprocessing.Rescaling(
+            scale=1./255., name="prestem_rescale"
+        )
```
sayakpaul, line 40: Where is the normalization layer? Or we are not using that anymore? I'm fine with either.

sayakpaul (changes requested):
Looks good @AdityaKane2001. 

I'd suggest running the scripts through `black` so that the linings and indentations come out a bit nicer.

regnety/config/config.py
AdityaKane2001: I'll remove the last two.

regnety/config/config.py
AdityaKane2001: Yes, we've metioned this on [L140](https://github.com/AdityaKane2001/regnety/blob/3baf92359754d67e89a84782a68213a3662cb32a/regnety/config/config.py#L140).

regnety/config/config.py
AdityaKane2001: I think it would be better to keep these names as they correspond to the ones in the paper.

> why is SE_ratio in such casing

No reason. I'll change that.

regnety/dataset/imagenet.py
AdityaKane2001: Sorry, but these are variable assignments.

regnety/models/blocks.py
AdityaKane2001: No we still have it on [L44](https://github.com/AdityaKane2001/regnety/blob/3baf92359754d67e89a84782a68213a3662cb32a/regnety/models/blocks.py#L44).

AdityaKane2001:
> I'd suggest running the scripts through black so that the linings and indentations come out a bit nicer.

Yes, I'll do that in a separate PR since reformatting will include (almost) the entire codebase in the diff.

AdityaKane2001:
@sayakpaul 

Made requested changes.

sayakpaul (approved):

AdityaKane2001:
Thanks for the approval, but we have to wait for Morgan as he had requested changes earlier.

AdityaKane2001:
@MorganR 

Could you please approve the PR?

regnety/evalaute.py
```diff
@@ -0,0 +1,79 @@
+import tensorflow as tf
```
MorganR, line 1: Should this file have been deleted in favour of regnety/evaluate.py? It looks like they are the same, and I'm guessing this one's name was a typo

MorganR (approved):
Just one minor change, but LGTM otherwise

regnety/evalaute.py
AdityaKane2001: Sorry, I was pushing from multiple endpoints. I'll check and remove that asap

AdityaKane2001:
@sayakpaul  @MorganR 

Thanks for your reviews and suggestions. Merging this to main now.

PR #20, opened 2021-08-14
Merge dev to main

AdityaKane2001:
Most files just have changes in `import` statement. For example, `regnety.regnety.config` is changed to `regnety.config`. Major changes in:
- imagenet.py
- blocks.py
- train.py
- train_utils.py


sayakpaul:
@AdityaKane2001 it's mentally very challenging to review changes made to 21 files. So, it'd be better if you could let us know about the particular scripts where you'd like our review.

AdityaKane2001:
@sayakpaul 

Sorry about that. I understand. This is happening since there are many tiny changes over the past few days.


Please review the files mentioned in the above [comment](https://github.com/AdityaKane2001/regnety/pull/20#issue-712713739). Here are the full file names and links:

| Filename                     | Purpose                | Link                                                                            |
|------------------------------|------------------------|---------------------------------------------------------------------------------|
| regnety/dataset/imagenet.py  | input pipeline         | [link](https://github.com/AdityaKane2001/regnety/blob/dev/regnety/dataset/imagenet.py ) |
| regnety/models/blocks.py     | model architecture     | [link](https://github.com/AdityaKane2001/regnety/blob/dev/regnety/utils/train_utils.py) |
| regnety/utils/train_utils.py | training configuration | [link](https://github.com/AdityaKane2001/regnety/blob/dev/regnety/utils/train_utils.py) |
| regnety/train.py             | training script        |  [link](https://github.com/AdityaKane2001/regnety/blob/dev/regnety/train.py)             |


/cc @MorganR

regnety/dataset/imagenet.py
```diff
@@ -149,7 +168,58 @@ def color_jitter(self, image: tf.Tensor, target: tf.Tensor) -> tuple:
...
 
         return aug_images, target
-
+    
+    def _inception_style_crop(self, images, labels):
+        """
+        Applies inception style cropping
+        """
+        # # Get target metrics
+        area_ratio = tf.random.uniform((), minval=0.08, maxval=1.0)
```
MorganR, line 177: `minval` here seems very small. Is this coming from the original code?

regnety/dataset/imagenet.py
```diff
@@ -149,7 +168,58 @@ def color_jitter(self, image: tf.Tensor, target: tf.Tensor) -> tuple:
...
+        y0s = tf.random.uniform((), minval=0, maxval=self.image_size - h + 1, dtype=tf.int32)
+        x0s = tf.random.uniform((), minval=0, maxval=self.image_size - w + 1, dtype=tf.int32)
+
+        begins = [0, y0s, x0s, 0]
+        sizes =  [self.batch_size, h, w, 3]
+
+        aug_images = tf.slice(images, begins, sizes)
+        aug_images = tf.image.resize(aug_images, (224, 224))
+        
+#         aug_images = tf.image.random_brightness(aug_images, max_delta=32. / 255.)
```
MorganR, line 198: Please delete if not needed

regnety/dataset/imagenet.py
```diff
@@ -149,7 +168,58 @@ def color_jitter(self, image: tf.Tensor, target: tf.Tensor) -> tuple:
...
+#         aug_images = tf.image.random_saturation(aug_images, lower=0.5, upper=1.5)
+
+        return aug_images, labels
+
+
+    def _pca_jitter(self, image, target):
```
MorganR, line 204: Please add documentation for this function.

regnety/models/utils.py
```diff
@@ -0,0 +1,37 @@
+import tensorflow as tf
```
MorganR, line 1: I assume this is written to parallel the behavior of the original regnet implementation. It would be great to add some module level documentation explaining this and referencing the original code.

Same for the _pca_jitter and _inception_style_crop functions from the other file.

regnety/train.py
```diff
@@ -5,60 +5,56 @@
...
-    default=None,
-)
-parser.add_argument(
-    "-tfp",
-    "--tfrecs_path_pattern",
-    type=str,
-    help="GCS bucket path pattern for tfrecords",
-)
+parser.add_argument("-taddr", "--tpu_address", type=str,
+                    help="Network address of TPU clsuter", default=None)
```
MorganR, line 26: nit: clsuter => cluster

regnety/train.py
```diff
@@ -5,60 +5,56 @@
...
-logging.basicConfig(
-    format="%(asctime)s %(levelname)s : %(message)s",
-    datefmt="%d-%b-%y %H:%M:%S",
-    level=logging.INFO,
-)
+logging.basicConfig(format="%(asctime)s %(levelname)s : %(message)s",
+                    datefmt="%d-%b-%y %H:%M:%S", level=logging.INFO)
 
-if "mf" not in flops:
+if ("mf" not in flops) or ("MF" not in flops):
```
MorganR, line 52: I think this should be an `and`, not an `or`

regnety/train.py
```diff
@@ -71,60 +67,61 @@
...
-)
+    f"Training on TFRecords: {train_prep_cfg.tfrecs_filepath[0]} to {train_prep_cfg.tfrecs_filepath[-1]}")
 logging.info(
-    f"Validating on TFRecords: {val_prep_cfg.tfrecs_filepath[0]} to {val_prep_cfg.tfrecs_filepath[-1]}"
-)
+    f"Validating on TFRecords: {val_prep_cfg.tfrecs_filepath[0]} to {val_prep_cfg.tfrecs_filepath[-1]}")
 
 with strategy.scope():
     model = tutil.make_model(flops, train_cfg)
+    # model.load_weights(log_location + "/init_weights/" + flops.upper())
```
MorganR, line 96: Delete?

MorganR (changes requested):

regnety/dataset/imagenet.py
AdityaKane2001: Yes, it is coming from the original code.

regnety/models/utils.py
AdityaKane2001: > I assume this is written to parallel the behavior of the original regnet implementation.


No, that is unintentional. I didn't find a better place to keep the initializers, so I made a utils file.  

I'll add the documentation.

regnety/train.py
AdityaKane2001: I'm not sure if we should remove this or not. We used to load random weights stored in a bucket. Should that be reflected in the final implementation?

/cc @sayakpaul

regnety/models/utils.py
AdityaKane2001: Added module level docstrings for important modules.

sayakpaul:
I comply with the changes suggested by @MorganR.

regnety/models/utils.py
MorganR: Sounds good. I'm not saying it looks the same (I haven't looked) I just know you were looking at that implementation to try to resolve the accuracy discrepancy.

regnety/train.py
MorganR: If there are changes in here that were not reflected in the trained models, it would be good to create and tag a branch with the code actually used to train version 1 on TF Hub.

That way you can delete any no-longer used code.

MorganR (approved):
Updated comments, otherwise LGTM

regnety/train.py
AdityaKane2001: Got it. In that case, I'll just uncomment this code. It was used in training.

AdityaKane2001:
@sayakpaul @MorganR 

Merging this one. Thank you for the reviews and suggestions.

PR #22, opened 2021-08-20
Added content to README and other minor changes

### LINKED ISSUE #15: A master table and other suggestions
@AdityaKane2001 

You have done an extremely complete and great job so far! Here's what I would suggest going further to make things a bit more manageable:

* Create a master table in the README with the following columns (you can change the names, of course): MF, Setup (keep the entires brief), Validation Top-1 Acc, Model Location. 
* Try working on a script to measure how many images per second a given RegNet is able to process on a commodity GPU like K80.

Cc: @MorganR

AdityaKane2001:
@sayakpaul @MorganR 

Requesting to review these files:
- [README.md](https://github.com/AdityaKane2001/regnety/blob/temp_readme/README.md): Added introduction, master table and usage
- [./regnety/models/model.py](https://github.com/AdityaKane2001/regnety/blob/temp_readme/regnety/models/model.py): added code for loading from GH release 

Minor changes in the notebook: 
- Added back softmax because without it results were incorrect for some reason. Also, it does not cause any confusion as we've specified that our model outputs logits. 
- Loading images using float32: `hub.KerasLayer` expects float tensors and not uint8.

Since these are small and functional changes, please ignore them for the review. 

/auto Closes #15


sayakpaul:
> README.md: Added introduction, master table and usage

For this one, I would suggest referring to your favorite repositories and notice how their README is structured and organized. This will force you to think back and come up with something that will cut through.  

> Added back softmax because without it results were incorrect for some reason.

It's quite strange then. Because calling `argmax()` on the logits should result in the same thing as calling `argmax()` on the softmaxed predictions. Will do the review of the model file in a bit.

regnety/models/model.py
```diff
@@ -76,4 +79,16 @@ def RegNetY(
 
     config = get_model_config(flops)
     model = _get_model_with_config(config)
+
+    if load_weights:
+        data_url = CKPT_URL + flops + "_best.data-00000-of-00001"
+        index_url = CKPT_URL + flops + "_best.index"
+
+        ckpt_path = tf.keras.utils.get_file(origin=data_url)
+        _ = tf.keras.utils.get_file(origin=index_url)
```
sayakpaul, line 88: Seems like it's not being used?

sayakpaul (changes requested):
Would also suggest making appropriate notes in the README so that the users are aware of the separate weights.

regnety/models/model.py
AdityaKane2001: Yes, we need the file but we have the downloaded file path in `ckpt_path` itself. So there's no need to store this path again.

AdityaKane2001:
@sayakpaul 

Could you please take a look at the [README](https://github.com/AdityaKane2001/regnety/blob/temp_readme/README.md) again? 

The notebook branch issue was resolved in [#72](https://github.com/tensorflow/tfhub.dev/pull/72).

sayakpaul:
Here are some thoughts:

* Your README introduction is lacking context. I think you should clarify what this repository is all about and that does not just mean it provides a TF 2.5 implementation of different variants of RegNetY. It should be a bit more expanded -- it's your GSoC project, it provides TPU-compatible training scipts for training on the ImageNet-1k dataset, checkpoints, etc. 
* I think your introduction could be titled as something "Note on RegNets". Then you could add stuff there. But I don't think "RegNetY is the most commonly used architecture for self supervised methods" is nearly true. ResNe50 is that architecture, believe it or not. So, refrain from asserting something like that. Instead, it could be more general, something like - "RegNets were proposed in [1] via a systematic model population search. The main purpose of RegNets is to [...] It has remarkable scaling capacities as shown in [2] (it could be the reference to the SEER paper). [...]"
* In the Y-Block diagram you could mention in the footnote that SE stands for Squeeze-Excitation.
* Before the checkpoint table, I would add an appropriate heading. I'd also make a note of the performance discrepancy there.
* For this [section](https://github.com/AdityaKane2001/regnety/tree/temp_readme#2-reloading-checkpoints-from-training), you might want to explain why the checkpoints have been hosted separately i.e. gradual unfreezing of layers, more fine-grained access etc. with a short code snippet. 
* Make the note on Colab Notebook explicitly clear and more visible. I'd consider putting this note at the top (repository introduction). 

All of these said I'd like to emphasize that READMEs should be created as per one's own taste. So, I encourage you to brainstorm more, consult a few different README use-cases, and decide for yourself what might be the best way of organizing information about the repository. This skill would go a really long way.

AdityaKane2001:
@sayakpaul 

Thanks for the in depth suggestions. I have tried to incorporate all of them. Please take a look at the [README](https://github.com/AdityaKane2001/regnety/blob/temp_readme/README.md) when you get a moment.

sayakpaul:
Looks good.

AdityaKane2001:
@sayakpaul 

If there's nothing more in the PR, could you please approve it?

/cc @MorganR

sayakpaul (approved):

sayakpaul:
With this I would like to request to remove the branch protection rule so that you can have a better grip at things.

AdityaKane2001:
@sayakpaul @MorganR

Merging this one. Thank you for your suggestions and reviews.

This was the last PR to be merged during GSoC. It has been a privilege to work and learn from you on this project.

